PR #2, opened 2015-01-16
Http add group

bosvos:


roledb.go
```diff
@@ -495,6 +495,34 @@ func (x *RoleGroupCache) InsertGroup(group *AuthGroup) (err error) {
 	return
 }
 
+func SaveGroup(icentral *Central, group *AuthGroup) bool {
```
benharper123: We don't need a SaveGroup function here, since it isn't adding any new functionality beyond what UpdateGroup already offers. Also, in Go one usually returns 'error' instead of a bool.

roledb.go
```diff
@@ -495,6 +495,34 @@ func (x *RoleGroupCache) InsertGroup(group *AuthGroup) (err error) {
...
+	if eupdate := icentral.GetRoleGroupDB().UpdateGroup(group); eupdate == nil {
+		return true
+	} else {
+		return false
+	}
+}
+
+func LoadOrCreateGroup(icentral *Central, groupName string, createIfNotExist bool) (*AuthGroup, error) {
```
benharper123: LoadOrCreateGroup should take a RoleGroupDB instead of a Central. Also, place the function directly beneath the definition of the RoleGroupDB interface, since this is a helper function that operates on anything that implements that interface.

bosvos:
Lets try again...

roledb.go
```diff
@@ -495,6 +495,26 @@ func (x *RoleGroupCache) InsertGroup(group *AuthGroup) (err error) {
 	return
 }
 
+func LoadOrCreateGroup(icentral *Central, groupName string, createIfNotExist bool) (*AuthGroup, error) {
```
benharper123: This function's signature should be:
func LoadOrCreateGroup(roleDB *RoleGroupDB, groupName string, createIfNotExist bool) (*AuthGroup, error)

This stuff is at a "lower" level that Central, and there shouldn't be any knowledge of Central down here.

Also, it would be best placed at line 95, because it is a function that operates on any object of the interface type RoleGroupDB.

bosvos: Done.

PR #6, opened 2016-02-10
Made changes to authaus, using userid as unique identifier instead of identity

theunissens:


bmharper:
In the unit tests, do not hardcode user ids. Fetch the user id from the username.

all_test.go
```diff
@@ -198,7 +208,7 @@ func TestRenameIdentity(t *testing.T) {
 	}
 
 	if _, err := c.GetTokenFromSession(session); err != ErrInvalidSessionToken {
-		t.Fatalf("All sessions for 'joe' should have been invalidated by rename")
+		t.Fatalf("All sessions for 'joe' should have been invalidated by rename %s", err)
```
bmharper: Leave some kind of punctuation between "rename" and "%s"

all_test.go
```diff
@@ -268,7 +278,8 @@ func TestResetPassword(t *testing.T) {
 func TestBasicAuth(t *testing.T) {
 	c := setup1(t)
 
-	expect_username_password := func(username, password, expectErrorStart, expectIdentity string) {
+	expect_username_password := func(username, password, expectErrorStart string, userId UserId) {
+		t.Log()
```
bmharper: t.Log() ?

all_test.go
```diff
@@ -268,7 +278,8 @@ func TestResetPassword(t *testing.T) {
 func TestBasicAuth(t *testing.T) {
 	c := setup1(t)
 
-	expect_username_password := func(username, password, expectErrorStart, expectIdentity string) {
+	expect_username_password := func(username, password, expectErrorStart string, userId UserId) {
```
bmharper: Words such as "expect", "truth", "actual" are very important in unit tests. Don't remove such words.

auth.go
```diff
@@ -51,6 +51,7 @@ func NewError(base error, detail string) error {
 // A Permit is an opaque binary string that encodes domain-specific roles.
 // This could be a string of bits with special meanings, or a blob of JSON, etc.
 type Permit struct {
+	name string
```
bmharper: What is Permit.name? It would need a very good reason to justify it's existence.

auth.go
```diff
@@ -86,7 +87,7 @@ the identity that performed the request, when this token expires, and
 the permit belonging to this identity.
 */
 type Token struct {
-	Identity string
+	UserId UserId
```
bmharper: This code has not been formatted with gofmt. You MUST use an editor that automatically runs gofmt on every save, otherwise you're just wasting your time (and others).

auth.go
```diff
@@ -205,6 +206,7 @@ type Central struct {
 	// on their operands.
 	Stats                  CentralStats
 	authenticator          Authenticator
+	userstore              UserStore
```
bmharper: Use consistent casing for the variable name "userStore". Here it is "userstore", and elsewhere "userStore". I suggest sticking to "userStore" everywhere.

auth.go
```diff
@@ -273,6 +282,10 @@ func NewCentralFromConfig(config *Config) (central *Central, err error) {
 		panic(err)
 	}
 
+	if userStore, err = NewUserStoreDB_SQL(&config.UserStore.DB); err != nil {
+		panic(err)
```
bmharper: Follow the convention here for NewPermitDB_SQL, which wraps up the error returned by NewPermitDB_SQL in a friendlier message.

auth.go
```diff
@@ -389,26 +407,35 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
...
 	token = &Token{}
-	if token.Identity, e = x.authenticator.Authenticate(identity, password); e == nil {
-		x.Log.Infof("Login authentication success (%v)", identity)
-		var permit *Permit
-		if permit, e = x.permitDB.GetPermit(identity); e == nil {
-			if x.MaxActiveSessions != 0 {
-				if e = x.sessionDB.InvalidateSessionsForIdentity(identity); e != nil {
-					x.Log.Errorf("Invalidate sessions for identity (%v) failed when enforcing MaxActiveSessions (%v)", identity, e)
-					return "", nil, e
+	if e = x.authenticator.Authenticate(identity, password); e == nil {
```
bmharper: Let's rather rewrite this function in the style:
```go
if x, e = something(); e != nil {
   return
}
if y, e = something(); e != nil {
    return
}
```
so that the code reads linearly, with the success case at the end.
You may also want to get rid of the named return values, if that adds any clarity, but that's hard to judge without actually writing the code. The important thing is that if an error is returned, then the return value MUST be `("", nil, err)`. I find that easier to guarantee when one doesn't have named return values.

auth.go
```diff
@@ -389,26 +407,35 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
...
 				}
-			}
-			token.Expires = time.Now().Add(x.NewSessionExpiresAfter)
-			token.Permit = *permit
-			sessionkey = generateSessionKey()
-			if e = x.sessionDB.Write(sessionkey, token); e == nil {
-				x.Stats.IncrementGoodLogin(x.Log)
-				x.Log.Infof("Login successful (%v)", identity)
-				return
+				x.Log.Infof("Permit (%v)", permit)
```
bmharper: Debugging log message left in.

auth.go
```diff
@@ -426,69 +453,69 @@ func (x *Central) Logout(sessionkey string) error {
...
-	x.Log.Infof("ResetPasswordFinish successful (%v)", identity)
-	x.sessionDB.InvalidateSessionsForIdentity(identity)
+	x.Log.Infof("ResetPasswordFinish successful (%v)", userId)
+	x.sessionDB.InvalidateSessionsForIdentity(userId)
 	return nil
 }
 
 // Create an identity in the Authenticator.
 // For the equivalent operation in the PermitDB, simply call SetPermit()
 func (x *Central) CreateAuthenticatorIdentity(identity, password string) error {
```
bmharper: This function is no longer accurately named. It should be called `CreateUserStoreIdentity`

auth.go
```diff
@@ -530,14 +557,19 @@ func (x *Central) RenameIdentity(oldIdent, newIdent string) error {
 
 	x.Log.Infof("RenameIdentity (PermitDB) successful (%v -> %v)", oldIdent, newIdent)
 
-	eInvalidate := x.InvalidateSessionsForIdentity(oldIdent)
-	x.Log.Infof("RenameIdentity (%v -> %v), session invalidation result (%v)", oldIdent, newIdent, eInvalidate)
-	return nil
+	if userId, eGetUserId := x.userstore.GetUserIdFromIdentity(newIdent); eGetUserId == nil {
```
bmharper: It is very important that a multistep process such as this, which is exposed as an apparently atomic API, tries it's best to be atomic. It's very bad if two parts of the process succeed, and the third part fails. I'm pretty sure that GetUserIdFromIdentity should be the first thing happening here. Also, we need to be checking whether the new identity is already in use, and failing with an appropriate error code if that is the case.

db.go
```diff
@@ -14,18 +14,30 @@ const (
 
 var (
 	veryFarFuture = time.Date(3000, 1, 1, 1, 1, 1, 1, time.UTC)
+	dummyUsers map[UserId]*dummyUser
```
bmharper: This may not be a global variable.

db.go
```diff
@@ -14,18 +14,30 @@ const (
...
 )
 
+type UserId uint64
```
bmharper: Stick to int64 instead. Almost universally, if you see a database that produces autoincrementing rowids, the type of that field is int64 and not uint64.

db.go
```diff
@@ -14,18 +14,30 @@ const (
...
 type Authenticator interface {
-	Authenticate(identity, password string) (string, error)                   // Return the non-canonicalized identity and nil error if the password is correct, otherwise one of ErrIdentityAuthNotFound or ErrInvalidPassword
-	SetPassword(identity, password string) error                              // This must not automatically create an identity if it does not already exist
-	ResetPasswordStart(identity string, expires time.Time) (string, error)    // Create a one-time token that can be used to reset the password with a subsequent call to ResetPasswordFinish
-	ResetPasswordFinish(identity string, token string, password string) error // Check that token matches the last one generated by ResetPasswordStart, and if so, call SetPassword
+	Authenticate(identity, password string) (error)                   	  // Return nil error if the password is correct, otherwise one of ErrIdentityAuthNotFound or ErrInvalidPassword
+	Close()                                                                   // Typically used to close a database handle
+}
+
+// The primary job of an authenticator is to validate an identity/password.
```
bmharper: You have copied and pasted comments here without reading or updating them.

db.go
```diff
@@ -14,18 +14,30 @@ const (
...
+// It can also be responsible for creating a new account.
+// All operations except for Close must be thread-safe.
+type UserStore interface {
+	SetPassword(userId UserId, password string) error                         // This must not automatically create an identity if it does not already exist
+	ResetPasswordStart(userId UserId, expires time.Time) (string, error)      // Create a one-time token that can be used to reset the password with a subsequent call to ResetPasswordFinish
+	ResetPasswordFinish(userId UserId, token string, password string) error   // Check that token matches the last one generated by ResetPasswordStart, and if so, call SetPassword
 	CreateIdentity(identity, password string) error                           // Create a new identity. If the identity already exists, then this must return ErrIdentityExists
 	RenameIdentity(oldIdent, newIdent string) error                           // Rename an identity. Returns ErrIdentityAuthNotFound if oldIdent does not exist. Returns ErrIdentityExists if newIdent already exists.
+	GetIdentityFromUserId(userId UserId) (string, error)			  // Gets the identity from the user id supplied
```
bmharper: Inconsistency in comments - "user id" vs "userid"

db.go
```diff
@@ -58,31 +70,46 @@ type SessionDB interface {
 
 // Authenticator that simply stores identity/passwords in memory
 type dummyAuthenticator struct {
-	users     map[string]*dummyUser
+	users     map[UserId]*dummyUser
```
bmharper: This is not the appropriate design for a dummy authenticator. The smoking gun is a *dummyUser present in both dummyUserStore and dummyAuthenticator. Those two ought to be different. The authenticator should store only passwords, and the userstore should store the same thing that a DB-based userstore stores.

db.go
```diff
@@ -127,48 +163,95 @@ func (x *dummyAuthenticator) ResetPasswordFinish(identity string, token string,
...
-func (x *dummyAuthenticator) CreateIdentity(identity, password string) error {
+func (x *dummyUserStore) CreateIdentity(identity, password string) error {
 	x.usersLock.Lock()
 	defer x.usersLock.Unlock()
-	if _, exists := x.users[CanonicalizeIdentity(identity)]; !exists {
-		x.users[CanonicalizeIdentity(identity)] = &dummyUser{identity, password, ""}
+	user := getUser(x.users, identity)
+	if user == nil {
+		userId := generateUserId(x.users)
+		x.users[userId] = &dummyUser{userId, identity, password, ""}
```
bmharper: You should canonicalize the identity before doing anything with it. In particular, here you are storing a non-canonicalized identity.

db.go
```diff
@@ -127,48 +163,95 @@ func (x *dummyAuthenticator) ResetPasswordFinish(identity string, token string,
...
 	x.usersLock.Lock()
 	defer x.usersLock.Unlock()
 
-	if _, exists := x.users[CanonicalizeIdentity(newIdent)]; exists {
-		return ErrIdentityExists
-	}
-
-	oldIdent = CanonicalizeIdentity(oldIdent)
 	newKey := CanonicalizeIdentity(newIdent)
+	newUser := getUser(x.users, newKey)
```
bmharper: Don't postpone `oldIdent = CanonicalizeIdentity(oldIdent)` until deeper into the function. This sort of thing is prone to producing bugs later one, perhaps with the 2nd or 3rd refactor. Instead, sanitize/canonicalize/etc your function arguments as soon as the function starts.

db.go
```diff
@@ -127,48 +163,95 @@ func (x *dummyAuthenticator) ResetPasswordFinish(identity string, token string,
...
+		oldIdent = CanonicalizeIdentity(oldIdent)
 
-	if user, exists := x.users[oldIdent]; exists {
-		delete(x.users, oldIdent)
-		x.users[newKey] = &dummyUser{newIdent, user.password, ""}
-		return nil
+		oldUser := getUser(x.users, oldIdent)
+
+		if oldUser != nil {
+			delete(x.users, oldUser.userId)
```
bmharper: Why delete and recreate? Why not just alter the name?

db.go
```diff
@@ -127,48 +163,95 @@ func (x *dummyAuthenticator) ResetPasswordFinish(identity string, token string,
...
+	for _, v := range x.users {
+		list = append(list, v.name)
 	}
 	x.usersLock.RUnlock()
 	return list, nil
 }
 
-func (x *dummyAuthenticator) Close() {
+func (x *dummyUserStore) GetIdentityFromUserId(userId UserId) (identity string, err error) {
+	x.usersLock.Lock()
```
bmharper: You don't need an exclusive lock here, since you're only reading

db.go
```diff
@@ -127,48 +163,95 @@ func (x *dummyAuthenticator) ResetPasswordFinish(identity string, token string,
...
+			identity = v.name
+			return
+		}
+	}
+	err = ErrIdentityAuthNotFound
+	return
+}
+
+func (x *dummyUserStore) GetUserIdFromIdentity(identity string) (userId UserId, err error) {
+	x.usersLock.Lock()
```
bmharper: Same comment here about locks.
Also, this is not an appropriate function for named return values.

db.go
```diff
@@ -127,48 +163,95 @@ func (x *dummyAuthenticator) ResetPasswordFinish(identity string, token string,
...
+	}
+	err = ErrIdentityAuthNotFound
+	return
+}
+
+func (x *dummyUserStore) Close() {
+}
+
+func getUser(users map[UserId]*dummyUser, identity string) (user *dummyUser) {
+	for _ , v := range users {
```
bmharper: No point in using named return values here.

db.go
```diff
@@ -127,48 +163,95 @@ func (x *dummyAuthenticator) ResetPasswordFinish(identity string, token string,
...
+		if CanonicalizeIdentity(v.name) == CanonicalizeIdentity(identity) {
+			user = v
+			return
+		}
+	}
+	return
+}
+
+func generateUserId(users map[UserId]*dummyUser) (UserId) {
```
bmharper: Return values should not be in parentheses if there is only one return value. I suspect gofmt would have fixed this up.
But this is a bad design - you are going to be reusing userids. Rather use an incrementing integer so that you never reuse old IDs.

db.go
```diff
@@ -178,18 +261,27 @@ type sanitizingAuthenticator struct {
 	backend Authenticator
 }
 
+// Authenticator that sanitizes inputs, so that we have more consistency with different backends
```
bmharper: The comment is invalid.

db.go
```diff
@@ -178,18 +261,27 @@ type sanitizingAuthenticator struct {
...
+	return strings.TrimSpace(password)
+}
+
 func cleanIdentityPassword(identity, password string) (string, string) {
-	return cleanIdentity(identity), strings.TrimSpace(password)
+	return cleanIdentity(identity), cleanPassword(password)
 }
 
-func (x *sanitizingAuthenticator) Authenticate(identity, password string) (string, error) {
+func (x *sanitizingAuthenticator) Authenticate(identity, password string) (error) {
```
bmharper: (error) -> error

db.go
```diff
@@ -198,34 +290,40 @@ func (x *sanitizingAuthenticator) Authenticate(identity, password string) (strin
...
+	if x.backend != nil {
+		x.backend.Close()
+		x.backend = nil
+	}
+}
+
+func (x *sanitizingUserStore) SetPassword(userId UserId, password string) error {
+	password = cleanPassword(password)
+	if userId <= 0 {
 		return ErrIdentityEmpty
```
bmharper: How can we know that userID <= 0 is an invalid userid?
Also, your original UserId was uint64, so testing for less than zero is pointless. If we absolutely need to, then I think it's OK to say that userID = 0 is an invalid userID, because I know of no databases that start generating autoincrementing IDs at zero (they all start at 1). However, don't make any such restriction unless it's necessary. In this case, there is no need. x.backend.SetPassword() will simply tell us that the ID is unknown, if  that is the case.

db.go
```diff
@@ -198,34 +290,40 @@ func (x *sanitizingAuthenticator) Authenticate(identity, password string) (strin
...
+func (x *sanitizingAuthenticator) Close() {
+	if x.backend != nil {
+		x.backend.Close()
+		x.backend = nil
+	}
+}
+
+func (x *sanitizingUserStore) SetPassword(userId UserId, password string) error {
+	password = cleanPassword(password)
+	if userId <= 0 {
```
bmharper: How can we know that userID <= 0 is an invalid userid?
Also, your original UserId was uint64, so testing for less than zero is pointless. If we absolutely need to, then I think it's OK to say that userID = 0 is an invalid userID, because I know of no databases that start generating autoincrementing IDs at zero (they all start at 1). However, don't make any such restriction unless it's necessary. In this case, there is no need. x.backend.SetPassword() will simply tell us that the ID is unknown, if that is the case.

db.go
```diff
@@ -198,34 +290,40 @@ func (x *sanitizingAuthenticator) Authenticate(identity, password string) (strin
...
 	}
-	return x.backend.SetPassword(identity, password)
+	return x.backend.SetPassword(userId, password)
 }
 
-func (x *sanitizingAuthenticator) ResetPasswordStart(identity string, expires time.Time) (string, error) {
-	identity = cleanIdentity(identity)
-	if len(identity) == 0 {
+func (x *sanitizingUserStore) ResetPasswordStart(userId UserId, expires time.Time) (string, error) {
+	if userId <= 0 {
```
bmharper: Same stuff here as above, and also further down.. same thing about userid == 0

db.go
```diff
@@ -560,33 +666,29 @@ func (x *dummyPermitDB) GetPermits() (map[string]*Permit, error) {
...
-		return ErrIdentityPermitNotFound
-	}
-	if _, newExists := x.permits[newIdent]; newExists {
-		return ErrIdentityExists
-	}
-
-	delete(x.permits, oldIdent)
-	x.permits[newIdent] = permit
-
+	// In this design, we store Authenticator and PermitDB in the same table, so we let the Authenticator portion
```
bmharper: Appropriate comment, but I think we ought to remove RenameIdentity from the PermitDB interface. I don't see how it  makes any sense now.

db.go
```diff
@@ -560,33 +666,29 @@ func (x *dummyPermitDB) GetPermits() (map[string]*Permit, error) {
...
+	// handle the rename. Once that's done, we have no work left to do here.
 	return nil
 }
 
 func (x *dummyPermitDB) Close() {
 }
+
+func getPermit(permits map[UserId]*Permit, identity string) (userId UserId, permit *Permit) {
```
bmharper: I believe this function is unused

db_sql.go
```diff
@@ -43,25 +44,42 @@ type sqlAuthenticationDB struct {
...
 }
 
-func (x *sqlAuthenticationDB) Authenticate(identity, password string) (id string, er error) {
-	row := x.db.QueryRow(`SELECT identity, password FROM authuser WHERE LOWER(identity) = $1`, CanonicalizeIdentity(identity))
-	dbHash := ""
-	id = identity
-	if err := row.Scan(&id, &dbHash); err != nil {
+func (x *sqlAuthenticationDB) Authenticate(identity, password string) (er error) {
+	//check identity exists
+	row := x.db.QueryRow(`SELECT userid FROM userstore WHERE LOWER(identity) = $1`, CanonicalizeIdentity(identity))
```
bmharper: Write this function in a linear fashion, the same as one of the earlier ones that I mentioned. On error, return, etc. No nesting on success.
Also - I don't think there's any need for the comment "check identity exists". It's pretty clear from "SELECT userid FROM userstore".

db_sql.go
```diff
@@ -133,31 +151,62 @@ func (x *sqlAuthenticationDB) ResetPasswordFinish(identity string, token string,
...
-		if _, ecreate := tx.Exec(`INSERT INTO authuser (identity, password) VALUES ($1, $2)`, identity, hash); ecreate == nil {
-			return tx.Commit()
+	//insert into user store
+	if tx1, etx1 := x.db.Begin(); etx1 == nil {
+		if _, ecreate1 := tx1.Exec(`INSERT INTO userstore (identity) VALUES ($1)`, identity); ecreate1 == nil {
+			tx1.Commit()
+
+			//get user id
+			var userId UserId
+			rows, err := x.db.Query(`SELECT userid FROM userstore WHERE identity = $1`, identity)
```
bmharper: Use `QueryRow` to make this code shorter.
This function must definitely be rewritten so that it's linear.
Also, you are starting transactions twice here. There is definitely no need for nested transactions here.

db_sql.go
```diff
@@ -195,7 +244,43 @@ func (x *sqlAuthenticationDB) GetIdentities() ([]string, error) {
 	return result, nil
 }
 
-func (x *sqlAuthenticationDB) Close() {
+func (x *sqlUserStoreDB) GetIdentityFromUserId(userId UserId) (identity string, err error) {
+	rows, err := x.db.Query(`SELECT identity FROM userstore WHERE userid = $1`, userId)
```
bmharper: Use QueryRow here

db_sql.go
```diff
@@ -195,7 +244,43 @@ func (x *sqlAuthenticationDB) GetIdentities() ([]string, error) {
...
+	}
+	if rows.Err() != nil {
+		err = rows.Err()
+		return
+	}
+	return
+}
+
+func (x *sqlUserStoreDB) GetUserIdFromIdentity(identity string) (userId UserId, err error) {
+	rows, err := x.db.Query(`SELECT userid FROM userstore WHERE identity = $1`, identity)
```
bmharper: Also, QueryRow here

db_sql.go
```diff
@@ -534,13 +632,19 @@ func SqlCreateSchema_Session(conx *DBConnection) error {
 	// we made some changes that are not backwards compatible, need to force a logout for all users
 	versions = append(versions, `DELETE FROM authsession;`)
 
+	versions = append(versions, `
+	DROP INDEX idx_authsession_identity;
+	ALTER TABLE authsession DROP COLUMN identity;
+	ALTER TABLE authsession ADD COLUMN userid BIGSERIAL;
+	ALTER TABLE authsession ADD CONSTRAINT fk_storeUserId_sessionUserId FOREIGN KEY (userid) REFERENCES userstore (userid);`)
```
bmharper: We should discuss the pros & cons of enabling these foreign key constraints. I'm not sure it's the right thing to do.
Also, there must be an explicit index on userid, so that we can efficiently wipe all sessions for a given user.

db_sql.go
```diff
@@ -553,5 +657,21 @@ func SqlCreateSchema_User(conx *DBConnection) error {
 	versions = append(versions, `
 	ALTER TABLE authuser ADD COLUMN pwdtoken VARCHAR;`)
 
+	versions = append(versions, `
+	DROP INDEX idx_authuser_identity;
```
bmharper: Since authuser already has a field "id BIGSERIAL", I think we should reuse that, but change it's name to "userid". Although Postgres does support multiple BIGSERIAL fields per table, it is the only database which supports this concept, so I'd rather not bank on that. If we need to drop tables and recreate them, that is fine, provided of course that we preserve existing user data. The easiest way to do this is to copy existing data to temporary tables, and then copy it back once the new structure is in place. All of that should be easy to do via simple SQL commands.
Also, we may need to use a different DB migration system, so that we can make changes to the different units (session, permit, auth, user) together. This is probably going to be necessary in order to preserve existing auth data.
Now might be a good time to switch to github.com\BurntSushi\migration for migrations. It's very similar in spirit to the existing system, but it is a bit more rigorous in certain aspects.

ldap.go
```diff
@@ -35,31 +34,43 @@ func (x *ldapAuthenticator) Authenticate(identity, password string) (id string,
...
+	//con *ldap.Conn
+	con *ldap.LDAPConnection
+}
+
+func (x *ldapUserStore) SetPassword(identity, password string) error {
 	return ErrUnsupported
 }
 
-func (x *ldapAuthenticator) ResetPasswordStart(identity string, expires time.Time) (string, error) {
+func (x *ldapUserStore) ResetPasswordStart(identity string, expires time.Time) (string, error) {
```
bmharper: I don't think we will ever support LDAP as a UserStore. For now, let's just implement LDAP as an Authenticator. This is the only real-world use case we've met so far.

bmharper:
If the review seems strict - it absolutely needs to be. Security code has to be as perfect as we can make it.

auth.go
theunissens: Looking at this, it is completely unused, I do not know why i never deleted it. It shouldnt be part of anything. Mistake on my part.

all_test.go
```diff
@@ -154,22 +155,23 @@ func isPrefix(prefix, str string) bool {
...
 
-	if e := c.CreateAuthenticatorIdentity("JOE", "123"); e == nil || !isPrefix(ErrIdentityExists.Error(), e.Error()) {
+	if e := c.CreateUserStoreIdentity("JOE", "123"); e == nil || !isPrefix(ErrIdentityExists.Error(), e.Error()) {
 		t.Errorf("CreateIdentity should fail because identities are case-insensitive. Instead, error is %v", e)
 	}
 	perm := Permit{}
 	roles := [2]byte{99}
 	perm.Roles = roles[:]
-	if e := c.SetPermit("JOE", &perm); e != nil {
+	if e := c.SetPermit(1, &perm); e != nil {
```
bmharper: These hardcoded user IDs should also be changed to symbols, the way it is done above, with variables such as joeUserID

auth.go
```diff
@@ -387,36 +402,49 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
...
+func (x *Central) Login(identity, password string) (string, *Token, error) {
+	token := &Token{}
+	if err := x.authenticator.Authenticate(identity, password); err != nil {
 		x.Stats.IncrementInvalidPasswords(x.Log)
-		x.Log.Infof("Login Authentication failed (%v) (%v)", identity, e)
+		x.Log.Infof("Login Authentication failed (%v) (%v)", identity, err)
+		return "", nil, err
+	}
+
+	var userId UserId
```
bmharper: It's customary for Go code such as this to be written in the following manner:
```go
userId, err := x.userStore.GetUserIdFromIdentity(identity)
if err != nil {
// ...
```
It is shorter this way.

auth.go
```diff
@@ -387,36 +402,49 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
...
+	var getUserIdErr error
+	if userId, getUserIdErr = x.userStore.GetUserIdFromIdentity(identity); getUserIdErr != nil {
+		x.Log.Errorf("UserId could not be found for identity (%v)", identity, getUserIdErr)
```
bmharper: The format string is missing a `%v`

auth.go
```diff
@@ -512,32 +540,28 @@ func (x *Central) RenameIdentity(oldIdent, newIdent string) error {
 		return nil
 	}
 
-	if err := x.authenticator.RenameIdentity(oldIdent, newIdent); err != nil {
-		x.Log.Infof("RenameIdentity failed (%v -> %v) (%v)", oldIdent, newIdent, err)
-		return err
+	var userId UserId
+	var eGetUserId error
+	if userId, eGetUserId = x.userStore.GetUserIdFromIdentity(oldIdent); eGetUserId != nil {
+		x.Log.Infof("Could not find userId for identity failed (%v) (%v)", oldIdent, ErrIdentityAuthNotFound)
```
bmharper: Typo - extra word "failed". Also, same comment here as above, about idomatic Go code and `if` statements.

db.go
```diff
@@ -16,73 +16,87 @@ var (
 	veryFarFuture = time.Date(3000, 1, 1, 1, 1, 1, 1, time.UTC)
 )
 
+type UserId int64
+
 // The primary job of an authenticator is to validate an identity/password.
 // It can also be responsible for creating a new account.
```
bmharper: Second line of function description is no longer true.

db.go
```diff
@@ -16,73 +16,87 @@ var (
...
 	name               string
 	password           string
 	passwordResetToken string
 }
 
-func newDummyAuthenticator() *dummyAuthenticator {
+func newDummyAuthenticator(dummyUserStore *dummyUserStore) *dummyAuthenticator {
 	d := &dummyAuthenticator{}
-	d.users = make(map[string]*dummyUser)
+	d.users = dummyUserStore.users
```
bmharper: This is a big no-no in maintaining sanity when writing multithreaded code. What's happening here is that you're disassociating the lock (dummyUserStore.usersLock) from the thing that it locks (dummyUserStore.users). That's opening up the door to bugs, because it's no longer clear what lock you need to take when reading or writing to 'users'.
What you should do instead here, is to make just a single struct, called dummyUserStoreAndAuth, which will act both as an authenticator, as well as a user store.

db.go
```diff
@@ -127,48 +150,93 @@ func (x *dummyAuthenticator) ResetPasswordFinish(identity string, token string,
...
 	}
 }
 
-func (x *dummyAuthenticator) GetIdentities() ([]string, error) {
+func (x *dummyUserStore) GetIdentities() ([]string, error) {
 	x.usersLock.RLock()
 	list := []string{}
-	for identity, _ := range x.users {
-		list = append(list, identity)
+	for _, v := range x.users {
```
bmharper: Stick to using `defer unlock`, because it's consistent with all the other member functions.

db.go
```diff
@@ -127,48 +150,93 @@ func (x *dummyAuthenticator) ResetPasswordFinish(identity string, token string,
...
+			return v.userId, nil
+		}
+	}
+	return -1, ErrIdentityAuthNotFound
+}
+
+func (x *dummyUserStore) Close() {
+}
+
+func getUser(users map[UserId]*dummyUser, identity string) *dummyUser {
```
bmharper: I'm pretty sure getUser and generateUserId are only used from methods of dummyUserStore, so they might as well be member functions.
Also - does it not complicate unit tests to have UserIds recycled? I think it would be more consistent with real-world database behaviour if UserIds were generated from a 64-bit integer that just always increments by one, and is only reset when the dummyUserStore object is discarded.

db_sql.go
```diff
@@ -133,17 +150,39 @@ func (x *sqlAuthenticationDB) ResetPasswordFinish(identity string, token string,
...
 }
 
-func (x *sqlAuthenticationDB) CreateIdentity(identity, password string) error {
+func (x *sqlUserStoreDB) CreateIdentity(identity, password string) error {
 	hash, ehash := computeAuthausHash(password)
 	if ehash != nil {
 		return ehash
 	}
 
+	//insert into user store
```
bmharper: For textual comments, leave a space after `//`
There are some other modifications which have the same spacing issue. You can find all such places with the following regex: `//[^\s]`

db_sql.go
```diff
@@ -133,17 +150,39 @@ func (x *sqlAuthenticationDB) ResetPasswordFinish(identity string, token string,
...
 	if tx, etx := x.db.Begin(); etx == nil {
-		if _, ecreate := tx.Exec(`INSERT INTO authuser (identity, password) VALUES ($1, $2)`, identity, hash); ecreate == nil {
+		if _, ecreate := tx.Exec(`INSERT INTO userstore (identity) VALUES ($1)`, identity); ecreate != nil {
+			//fmt.Printf("CreateIdentity failed because: %v", ecreate)
```
bmharper: Everything inside this function must be wrapped inside a single transaction, and if any step fails, then the whole transaction must roll back. There is zero point in wrapping just the `INSERT INTO userstore` inside a transaction, because single statements are by definition always wrapped inside a transaction.

db_sql.go
```diff
@@ -469,89 +539,155 @@ func SqlCreateDatabase(conx *DBConnection) error {
...
+		`CREATE TABLE authuser (id BIGSERIAL PRIMARY KEY, identity VARCHAR, password VARCHAR, permit VARCHAR);
+		CREATE UNIQUE INDEX idx_authuser_identity ON authuser (identity);`,
+
+		`DROP INDEX idx_authuser_identity;
+		CREATE UNIQUE INDEX idx_authuser_identity ON authuser (LOWER(identity));`,
+
+		`ALTER TABLE authuser ADD COLUMN pwdtoken VARCHAR;`,
+
+		//Change from using identity as unique reference to user account to using userid
+		`CREATE TABLE userstore (userid BIGSERIAL PRIMARY KEY, identity VARCHAR);
```
bmharper: Since all other tables are prefixed with 'auth', I think userstore should also be called authuserstore.
I would also change the old table 'authuser' to 'authuserpwd', or something indicating that this is responsible for verification.

The comment here should be a bit clearer, such as "Change from using email address as the primary identity of a user, to a 64-bit integer, which we call UserId".
And also, the comment should mention that the above migrations were from an old system, and we're bootstrapping them into the new burntsushi migration system.

Then.. I think it's much better to drop the old tables and recreate them, vs altering them to suit their new role. Since all existing data is getting saved into _temp tables, there is no virtue in incrementally altering the tables.
The downside to altering the tables in this manner, is that it's very hard for a reader to see what the current schema is, because he needs to parse all the old migrations, as well as these changes.
Far better to just have plain old "CREATE TABLE ...." statements in there.

Another thing - shouldn't identity be called 'email' here?

Also - it's highly inadvisable to create a BIGSERIAL field, and then insert explicit values into it. As we discussed, this ends up leaving the sequence generator behind the reality of what's inside the table, and the next time you try to insert an auto-generated ID, you'll get a key violation. So don't mix and match auto-generated IDs with explicit IDs. As an example of where that's happening, see the insert into authsession: the userid field is BIGSERIAL.

Another thing - why `DROP TABLE IF EXISTS authsession_temp`? There is zero ambiguity here - authsession_temp HAS to exist, so rather just write DROP TABLE, because the IF EXISTS introduces a ton of self-doubt.

db_sql.go
```diff
@@ -469,89 +539,155 @@ func SqlCreateDatabase(conx *DBConnection) error {
...
+		JOIN userstore AS us ON us.identity = authtemp.identity;
+		DROP TABLE IF EXISTS authsession_temp;
+
+		CREATE TABLE authuser_temp(id BIGSERIAL PRIMARY KEY, identity VARCHAR, password VARCHAR, permit VARCHAR, pwdtoken VARCHAR);
+		INSERT INTO authuser_temp (id, identity, password, permit, pwdtoken) SELECT * FROM authuser;
+		TRUNCATE authuser;
+		DROP INDEX idx_authuser_identity;
+		ALTER TABLE authuser DROP COLUMN identity;
+		ALTER TABLE authuser ADD COLUMN userid BIGSERIAL;
+		ALTER TABLE authuser ADD CONSTRAINT fk_storeUserId_authUserId FOREIGN KEY (userid) REFERENCES userstore (userid);
```
bmharper: As we discussed, we don't want foreign key constraints in here.

db_sql.go
```diff
@@ -469,89 +539,155 @@ func SqlCreateDatabase(conx *DBConnection) error {
...
+		migrations = append(migrations, func(tx migration.LimitedTx) error {
+			_, err := tx.Exec(srcCapture)
+			return err
+		})
+	}
+	return migrations
+}
+
+//This function makes sure that the old migrations that were part of the old migration tool have run.
+func runBootstrap(db *sql.DB) (bool, error) {
```
bmharper: The documentation of this function doesn't specify what the return value means.

db_sql.go
```diff
@@ -469,89 +539,155 @@ func SqlCreateDatabase(conx *DBConnection) error {
...
+	oldVersionTables := []string{"authgroup_version", "authsession_version", "authuser_version"}
+	// We first want to see if the new migrator has taken over, by checking if old version tracking tables exist
+	tableNameResult := ""
+	query := db.QueryRow(fmt.Sprintf("SELECT table_name FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = '%v'", oldVersionTables[0]))
+	err := query.Scan(&tableNameResult)
+	if err != nil {
+		if strings.Index(err.Error(), "no rows in result set") != -1 {
```
bmharper: I'm skeptical on there being two "up to date" return paths. Under which conditions will one hit the first `return true, nil` branch, and under which conditions will one hit the second such branch?

db_sql.go
```diff
@@ -469,89 +539,155 @@ func SqlCreateDatabase(conx *DBConnection) error {
...
-	versions := make([]string, 0)
-	versions = append(versions, `
-	CREATE TABLE authsession (id BIGSERIAL PRIMARY KEY, sessionkey VARCHAR, identity VARCHAR, permit VARCHAR, expires TIMESTAMP);
-	CREATE UNIQUE INDEX idx_authsession_token    ON authsession (sessionkey);
-	CREATE        INDEX idx_authsession_identity ON authsession (identity);
-	CREATE        INDEX idx_authsession_expires  ON authsession (expires);`)
-	// we made some changes that are not backwards compatible, need to force a logout for all users
-	versions = append(versions, `DELETE FROM authsession;`)
+	//
+	checkTableVersion := func(table string, expectedVersion int) bool {
```
bmharper: `check` is usually a bad prefix for a function that returns a bool. You never know whether the return value means `OK` or `Not OK`. In a case like this, I would rather just make a function such as "getTableVersion" that returns (int, error). 
It's also not clear here that a DB-level error is being treated correctly. For example, if the DB connection dies on one of these queries, then checkVersion will return false, which will end up causing the system to try and upgrade the DB schema. This is not right - if one receives an error from the DB, then everything must abort.

db_sql.go
```diff
@@ -469,89 +539,155 @@ func SqlCreateDatabase(conx *DBConnection) error {
...
+			_, err := db.Exec(fmt.Sprintf("DROP TABLE IF EXISTS %v", oldVersionTable))
+			if err != nil {
+				return false, err
+			}
+		}
+	}
 
-	versions = append(versions, `
-	ALTER TABLE authuser ADD COLUMN pwdtoken VARCHAR;`)
+	// Now we need to set the version in new migration version tracking table
```
bmharper: Generally avoid starting a comment with "Now....". It's generally a statement by a programmer of the code that needs to be written next, but a comment should read as though it's documenting the final product.
Also - no need to apologise for the way we're bootstrapping the burntsushi migration system. It's the only reasonable thing to do. We certainly do need an explanation in here, but don't make it sound as though there's a better solution.

db_sql.go
```diff
@@ -469,89 +539,155 @@ func SqlCreateDatabase(conx *DBConnection) error {
...
-	return migrateSchemaInternal(true, conx, schema_name, migrations)
-}
-
-func migrateSchemaInternal(tryCreatingDB bool, conx *DBConnection, schema_name string, migrations []string) (migrateError error) {
-	if db, eConnect := conx.Connect(); eConnect != nil {
-		return eConnect
-	} else {
-		defer db.Close()
+func runMigrations(db *sql.DB, conx *DBConnection) (bool, error) {
+	// We have started using a migration tool by BurntSushi. This requires us to delete old version tracking tables etc.
```
bmharper: Avoid writing comments that pretend as though the only time is the present. When somebody reads "We have started using..." he has no idea when that happened. Rather phrase it as though you're writing history, for example:
"Authaus initially used it's own migration system, but in January 2016, we moved away from that, and instead started using the Burntsushi migration system. In order to make that change, we needed to...."

PR #7, opened 2016-03-11
UserId as unique identifier instead of identity. Added extra user fields with.

theunissens:


bmharper:
I assume the commit message "Added extra user fields with ability to lock, unlock and archive a user" is out of date. Presumably it is "Added extra user fields with ability to archive a user".
Also, on the first commit message - it's tautological to say "Made changes to authaus." This entire repo IS authaus, so you can't possibly be changing anything else.

bmharper:
It looks like a lot of the changes in the second commit actually belong in the first commit. Can you squash them first please, and then I'll take a look?
This seems like the kind of big change that really just deserves a single commit. One doesn't want to track a single event in history over two commits - that makes it very hard to analyze in retrospect.

all_test.go
```diff
@@ -32,6 +32,11 @@ and Postgres backend. I'm not sure that testing without the race detector adds a
 */
 
 var backend_postgres = flag.Bool("backend_postgres", false, "Run tests against Postgres backend")
+var joeUserId UserId = 1
```
bmharper: I guess it's OK if we leave these hardcoded, but then at least let's leave a comment here saying `These hard-coded UserId values are predictable because we always drop & recreate the postgres backend when running tests, so we know that our IDs always start at 1`. Reading code with magic constants is very annoying if they aren't explained.

all_test.go
```diff
@@ -154,24 +166,24 @@ func isPrefix(prefix, str string) bool {
...
 
-	if e := c.CreateAuthenticatorIdentity("JOE", "123"); e == nil || !isPrefix(ErrIdentityExists.Error(), e.Error()) {
+	if e := c.CreateUserStoreIdentity("JOE", "JOEusername", "JOEfirstname", "JOElastname", "JOE084", "123"); e == nil || !isPrefix(ErrIdentityExists.Error(), e.Error()) {
 		t.Errorf("CreateIdentity should fail because identities are case-insensitive. Instead, error is %v", e)
 	}
 	perm := Permit{}
 	roles := [2]byte{99}
 	perm.Roles = roles[:]
-	if e := c.SetPermit("JOE", &perm); e != nil {
+	if e := c.SetPermit(joeUserId, &perm); e != nil {
```
bmharper: This part of the test, that checks case sensitivity of Permits, is no longer valid, since you're now passing those APIs integer IDs.

all_test.go
```diff
@@ -198,31 +210,33 @@ func TestRenameIdentity(t *testing.T) {
...
 	}
+
+	teardown(c)
 }
 
 func TestResetPassword(t *testing.T) {
 	c := setup1(t)
 
-	if _, err := c.ResetPasswordStart("nobody_ever", time.Now()); err != ErrIdentityAuthNotFound {
+	if _, err := c.ResetPasswordStart(999, time.Now()); err != ErrIdentityAuthNotFound {
```
bmharper: I think it's OK to either sprinkle `999` throughout the code, or always use the constant `notFoundUserId`, but you shouldn't mix the two.

all_test.go
```diff
@@ -198,31 +210,33 @@ func TestRenameIdentity(t *testing.T) {
...
 		t.Fatalf("ResetPasswordStart should fail with ErrIdentityAuthNotFound instead of %v", err)
 	}
-	if err := c.ResetPasswordFinish("joe", "", "12345"); err != ErrInvalidPasswordToken {
+	if err := c.ResetPasswordFinish(1, "", "12345"); err != ErrInvalidPasswordToken {
```
bmharper: Surely the `1` here should be `joeUserId`.

all_test.go
```diff
@@ -259,16 +273,19 @@ func TestResetPassword(t *testing.T) {
...
 	}
+
+	teardown(c)
 }
 
 func TestBasicAuth(t *testing.T) {
 	c := setup1(t)
 
-	expect_username_password := func(username, password, expectErrorStart, expectIdentity string) {
+	expect_username_password := func(username, password, expectErrorStart string, expectIdentity string) {
```
bmharper: No need to make this change to the function definition. The original version was correct, and idiomatic.

all_test.go
```diff
@@ -259,16 +273,19 @@ func TestResetPassword(t *testing.T) {
...
+		//authenticate
```
bmharper: Leave a single space after `//`

all_test.go
```diff
@@ -364,10 +390,10 @@ func TestPermitChange(t *testing.T) {
 	perm2.Roles = perm2_roles[:]
 	for nsessions := 1; nsessions < 100; nsessions *= 2 {
 		// restore password and permit
-		if e := c.SetPassword("joe", "123"); e != nil {
+		if e := c.SetPassword(1, "123"); e != nil {
```
bmharper: Surely `1` should be `joeUserId`?

all_test.go
```diff
@@ -500,4 +534,80 @@ func TestSessionDelete(t *testing.T) {
...
+func TestUpdateIdentity(t *testing.T) {
+	c := setup1(t)
+
+	newEmail := "newEmail"
+	newUsername := "newUsername"
+	newName := "newName"
+	newSurname := "newSurname"
+	newMobile := "newMobile"
+
+	if err := c.UpdateIdentity(notFoundUserId, newEmail, newUsername, newName, newSurname, newMobile); err != nil {
```
bmharper: The logic is hard to follow here. It would be clearer if it read:
```go
if err := c.UpdateIdentity(...); err != ErrIdentityAuthNotFound {
    t.Fatalf("Expected ErrIdentityAuthNotFound, but instead got %v", err)
}
```

all_test.go
```diff
@@ -500,4 +534,80 @@ func TestSessionDelete(t *testing.T) {
...
+		if err != ErrIdentityAuthNotFound {
+			t.Fatalf("TestUpdateIdentity failed: an error occurred: %v", err)
+		}
+	} else {
+		t.Fatalf("Update should not have succeeded")
+	}
+
+	if err := c.UpdateIdentity(joeUserId, newEmail, newUsername, newName, newSurname, newMobile); err != nil {
```
bmharper: Print the error message here

all_test.go
```diff
@@ -500,4 +534,80 @@ func TestSessionDelete(t *testing.T) {
...
+		t.Fatalf("Update should not have failed")
+	}
+
+	users, e := c.GetAuthenticatorIdentities()
```
bmharper: If you start using `err` in a function, then stick to `err`. Don't switch arbitrarily inside a single function between `err` and `e`.

all_test.go
```diff
@@ -500,4 +534,80 @@ func TestSessionDelete(t *testing.T) {
...
+	if e != nil {
+		t.Fatalf("TestUpdateIdentity failed: $v", e)
```
bmharper: $v is not correct.

all_test.go
```diff
@@ -500,4 +534,80 @@ func TestSessionDelete(t *testing.T) {
...
+
+	updateSuccess := false
+	for _, user := range users {
+		if user.UserId == joeUserId && CanonicalizeIdentity(user.Email) == CanonicalizeIdentity(newEmail) && user.Username == newUsername && user.Firstname == newName && user.Lastname == newSurname && user.Mobilenumber == newMobile {
+			updateSuccess = true
+			break
+		}
+	}
+	if !updateSuccess {
+		t.Fatalf("TestUpdateIdentity failed")
```
bmharper: It would be better here have an information error message such as "After update, failed to find updated user"

all_test.go
```diff
@@ -500,4 +534,80 @@ func TestSessionDelete(t *testing.T) {
...
+	}
+
+	teardown(c)
+}
+
+func TestArchiveIdentity(t *testing.T) {
+	c := setup1(t)
+
+	if err := c.ArchiveIdentity(notFoundUserId); err != nil {
```
bmharper: Same comment here as `TestUpdateIdentity`, regarding logic flow clarity.

all_test.go
```diff
@@ -500,4 +534,80 @@ func TestSessionDelete(t *testing.T) {
...
+		if err != ErrIdentityAuthNotFound {
+			t.Fatalf("TestArchiveIdentity failed: an error occurred: %v", err)
+		}
+	} else {
+		t.Fatalf("Archive should not have succeeded")
+	}
+
+	if err := c.ArchiveIdentity(joeUserId); err != nil {
```
bmharper: Many of the same comments here, as in the previous function.

auth.go
```diff
@@ -387,36 +404,49 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
...
-			if e = x.sessionDB.Write(sessionkey, token); e == nil {
-				x.Stats.IncrementGoodLogin(x.Log)
-				x.Log.Infof("Login successful (%v)", identity)
-				return
-			}
-		} else {
-			x.Log.Infof("Login GetPermit failed (%v) (%v)", identity, e)
-		}
-	} else {
+func (x *Central) Login(identity, password string) (string, *Token, error) {
```
bmharper: In a function definition such as this, you _must_ name the return values, because it's not at all obvious to a reader of this function definition, what the first "string" return value holds. It's an eternal rule of any function that if you need to read it's body to determine what it does, then you've failed to create a good API. As a litmus test, you can keep "godoc" running, and make sure that you can effectively read an API just by looking at the godoc page that is generated for that API.

auth.go
```diff
@@ -426,77 +456,120 @@ func (x *Central) Logout(sessionkey string) error {
...
-// For the equivalent operation in the PermitDB, simply call SetPermit()
-func (x *Central) CreateAuthenticatorIdentity(identity, password string) error {
-	e := x.authenticator.CreateIdentity(identity, password)
+// Create an identity in the AuthUserStore.
+func (x *Central) CreateUserStoreIdentity(email, username, firstname, lastname, mobilenumber, password string) error {
+	e := x.userStore.CreateIdentity(email, username, firstname, lastname, mobilenumber, password)
+	if e == nil {
+		x.Log.Infof("CreateAuthenticatorIdentity successful (%v)", email)
+	} else {
+		x.Log.Infof("CreateAuthenticatorIdentity failed (%v) (%v)", email, e)
```
bmharper: Failure here should be an 'Error', and not an 'Info' message, because new users will always be created by an 'admin' user, so we expect this to succeed.
Also, it would probably be good to include the UserId in the success message. From here on out, most of the log messages will refer to the UserId and not the email, so this is the only point where one can sync the two up, if all you have is the logs. Perhaps `CreateIdentity` should return `(UserId, error)`?

auth.go
```diff
@@ -426,77 +456,120 @@ func (x *Central) Logout(sessionkey string) error {
...
 	return e
 }
 
+// Archive a user in the AuthUserStore.
+func (x *Central) ArchiveIdentity(userId UserId) error {
+	e := x.userStore.ArchiveIdentity(userId)
+	if e == nil {
+		x.Log.Infof("Archive Identity successful (%v)", userId)
+	} else {
+		x.Log.Infof("Archive Identity failed (%v) (%v)", userId, e)
```
bmharper: Same comment here - failure should be Error not Info.

auth.go
```diff
@@ -426,77 +456,120 @@ func (x *Central) Logout(sessionkey string) error {
...
+
+// Get userid from identity.
+func (x *Central) GetUserIdFromIdentity(identity string) (UserId, error) {
+	userId, e := x.userStore.GetUserIdFromIdentity(identity)
+	if e == nil {
+		return userId, nil
+	} else {
+		x.Log.Infof("GetUserIdFromIdentity failed (%v) (%v)", identity, e)
+	}
+	return -1, e
```
bmharper: We need a constant NullUserId, which we should return here:
`const NullUserId UserId = 0`
Unless you have constraints which dictates otherwise, you should always prefer zero over -1 for a null value. The reason is that 0 is the default value for an integer in Go. The same applies to C, because there you want to be able to use memset(0) to initialize a data structure. Also, I have never met a database that will return 0 as it's first autogenerated ID, so 0 should be the de-facto null value for integers.

auth.go
```diff
@@ -512,32 +585,27 @@ func (x *Central) RenameIdentity(oldIdent, newIdent string) error {
...
-	x.Log.Infof("RenameIdentity (Authenticator) successful (%v -> %v)", oldIdent, newIdent)
-
-	if err := x.permitDB.RenameIdentity(oldIdent, newIdent); err != nil {
-		x.Log.Warnf("RenameIdentity (PermitDB) failed (%v -> %v) (%v)", oldIdent, newIdent, err)
-
-		errReverse := x.authenticator.RenameIdentity(newIdent, oldIdent)
-		x.Log.Infof("Reverse of rename (Authenticator) (%v -> %v). Result (%v)", newIdent, oldIdent, errReverse)
-
+	if err := x.userStore.RenameIdentity(oldIdent, newIdent); err != nil {
+		x.Log.Infof("RenameIdentity failed (%v -> %v) (%v)", oldIdent, newIdent, err)
```
bmharper: I think this should be a warning, not info.

db.go
```diff
@@ -127,48 +154,119 @@ func (x *dummyAuthenticator) ResetPasswordFinish(identity string, token string,
...
+
+	return "", ErrIdentityAuthNotFound
+}
+
+func (x *dummyUserStoreAndAuth) GetUserIdFromIdentity(identity string) (UserId, error) {
+	x.usersLock.RLock()
+	defer x.usersLock.RUnlock()
+
+	if CanonicalizeIdentity(identity) == "" {
+		return -1, ErrIdentityEmpty
```
bmharper: Use NullUserId

db.go
```diff
@@ -127,48 +154,119 @@ func (x *dummyAuthenticator) ResetPasswordFinish(identity string, token string,
...
+
+	for _, v := range x.users {
+		if v.email == identity {
+			return v.userId, nil
+		}
+	}
+	return -1, ErrIdentityAuthNotFound
+}
+
+func getUser(users map[UserId]*dummyUser, email string) *dummyUser {
```
bmharper: Since `getUser` function only ever operates on members of dummyUserStoreAndAuth, I would make it a member of dummyUserStoreAndAuth. This is just to narrow the scope of things.
Same goes for `generateUserId`.

db.go
```diff
@@ -127,48 +154,119 @@ func (x *dummyAuthenticator) ResetPasswordFinish(identity string, token string,
...
+	for _, v := range users {
+		if CanonicalizeIdentity(v.email) == CanonicalizeIdentity(email) {
+			return v
+		}
+	}
+	return nil
+}
+
+func generateUserId(users map[UserId]*dummyUser) UserId {
+	return UserId(len(users) + 1)
```
bmharper: It would be far better here to mimick the technique used by real databases, which is to keep a `nextUserId` variable, and always return that value, then increment it by one, whenever creating a new user. I know this is only used for unit tests, but it's the right thing to do. Some day, somebody is going to waste an hour tracking down a strange unit test failure, when it would have been 1 minute with an incrementing counter. The technique used here is prone to generating the same ID twice, which is not something one would expect. The two different methods use almost the same number of code lines, so might as well use the more robust one.

db_sql.go
```diff
@@ -133,31 +150,84 @@ func (x *sqlAuthenticationDB) ResetPasswordFinish(identity string, token string,
...
+			}
+			tx.Rollback()
+			return eCreateUserStore
+		}
+
+		// Get user id
+		var userId int64
+		row := tx.QueryRow(`SELECT userid FROM authuserstore WHERE email = $1`, email)
+		if scanErr := row.Scan(&userId); scanErr != nil {
+			return scanErr
```
bmharper: Missing a rollback on this error condition.

db_sql.go
```diff
@@ -43,25 +45,40 @@ type sqlAuthenticationDB struct {
 	db *sql.DB
 }
 
-func (x *sqlAuthenticationDB) Authenticate(identity, password string) (id string, er error) {
-	row := x.db.QueryRow(`SELECT identity, password FROM authuser WHERE LOWER(identity) = $1`, CanonicalizeIdentity(identity))
+func (x *sqlAuthenticationDB) Authenticate(identity, password string) error {
+	row := x.db.QueryRow(`SELECT userid FROM authuserstore WHERE LOWER(email) = $1`, CanonicalizeIdentity(identity))
```
bmharper: Should Authenticate not fail if a user is archived?

all_test.go
```diff
@@ -500,4 +534,80 @@ func TestSessionDelete(t *testing.T) {
...
+		t.Fatalf("Archive should not have failed")
+	}
+
```
bmharper: What exactly is the definition of Archive? Surely a user should fail to authenticate after he's been archived. What about changing passwords after you've been archived, or renaming, or updating?

auth.go
```diff
@@ -426,77 +456,120 @@ func (x *Central) Logout(sessionkey string) error {
...
+		x.Log.Infof("Update Identity successful (%v)", userId)
 	} else {
-		x.Log.Infof("CreateAuthenticatorIdentity failed (%v) (%v)", identity, e)
+		x.Log.Infof("Update Identity failed (%v) (%v)", userId, e)
 	}
 	return e
 }
 
+// Archive a user in the AuthUserStore.
+func (x *Central) ArchiveIdentity(userId UserId) error {
```
bmharper: Surely we should be invalidating all existing sessions for this user here?

db_sql.go
```diff
@@ -469,89 +580,161 @@ func SqlCreateDatabase(conx *DBConnection) error {
 	}
 }
 
-func MigrateSchema(conx *DBConnection, schema_name string, migrations []string) error {
-	return migrateSchemaInternal(true, conx, schema_name, migrations)
-}
+func RunMigrations(conx *DBConnection) (bool, error) {
+	// Authaus used its own migration tool, but we moved away from that and started using a migration tool by BurntSushi. This requires us to delete old version tracking tables etc.
```
bmharper: This comment like this should include a genuine date. For example,
"Until March 2016, Authaus used it's own migration tool...."

db_sql.go
```diff
@@ -469,89 +580,161 @@ func SqlCreateDatabase(conx *DBConnection) error {
...
+
+		`DROP INDEX idx_authuser_identity;
+		CREATE UNIQUE INDEX idx_authuser_identity ON authuser (LOWER(identity));`,
+
+		`ALTER TABLE authuser ADD COLUMN pwdtoken VARCHAR;`,
+
+		// Change from using email address as the primary identity of a user, to a 64-bit integer, which we call UserId.
+		// The above migrations were from the old migration system, from this migration on we are using the new migration
+		// system from BurntSushi
+		`CREATE TABLE authuserstore (userid BIGSERIAL PRIMARY KEY, email VARCHAR, username VARCHAR, firstname VARCHAR, lastname VARCHAR, mobile VARCHAR, archived BOOLEAN default false);
```
bmharper: I don't think we should rely on DEFAULT db values (ie used here for the archived field). We should rather be explicit. It's like static type checking. If we don't have any DEFAULT value, then we will see a NULL value for archived, and know that we've forgotten to set it somewhere.

db_sql.go
```diff
@@ -469,89 +580,161 @@ func SqlCreateDatabase(conx *DBConnection) error {
...
+		migrations = append(migrations, func(tx migration.LimitedTx) error {
+			_, err := tx.Exec(srcCapture)
+			return err
+		})
+	}
+	return migrations
+}
+
+// This function makes sure that the old migrations that were part of the old migration tool have run, returns true on success else false with error
+func runBootstrap(conx *DBConnection) (bool, error) {
```
bmharper: This function shouldn't return (bool, error), but simply error.

db_sql.go
```diff
@@ -469,89 +580,161 @@ func SqlCreateDatabase(conx *DBConnection) error {
...
+	}
+	version3, err3 := getTableVersion(oldVersionTables[2])
+	if err3 != nil {
+		return false, err3
+	}
 
-	versions = append(versions, `
-	DROP INDEX idx_authuser_identity;
-	CREATE UNIQUE INDEX idx_authuser_identity ON authuser (LOWER(identity));`)
+	if version1 == 1 && version2 == 2 && version3 == 3 {
```
bmharper: There should be a comment here explaining the magic constants 1, 2, 3. To the casual observer, it looks strange that they're perfectly correlated with version1, version2, version3

bmharper:
Aside from the all the little things, I'm a little worried about the lack of rigor regarding the "archived" property. This hasn't been well defined, so we should clarify what exactly it means to be "archived".

all_test.go
```diff
@@ -364,10 +388,10 @@ func TestPermitChange(t *testing.T) {
 	perm2.Roles = perm2_roles[:]
 	for nsessions := 1; nsessions < 100; nsessions *= 2 {
 		// restore password and permit
-		if e := c.SetPassword("joe", "123"); e != nil {
+		if e := c.SetPassword(joeUserId, "123"); e != nil {
 			t.Fatalf("Password restore failed: %v (nsessions = %v)", e, nsessions)
 		}
-		if e := c.SetPermit("joe", &perm1); e != nil {
+		if e := c.SetPermit(1, &perm1); e != nil {
```
bmharper: This `1` should be `joeUserId`

all_test.go
```diff
@@ -259,25 +271,31 @@ func TestResetPassword(t *testing.T) {
...
 	}
+
+	teardown(c)
 }
 
 func TestBasicAuth(t *testing.T) {
 	c := setup1(t)
 
 	expect_username_password := func(username, password, expectErrorStart, expectIdentity string) {
+		//authenticate
```
bmharper: We always leave a single space after `//`. I did mention this in the previous round of comments. The previous issue was also mentioned in the previous round of comments.

bmharper:
Please go meticulously through every comment that I made on the previous round. I only checked now up to the first 12 issues that I previously commented on, and 3 out of those first 12 haven't been attended to. I know that some of them are pretty minor issues, such as formatting, or symbolic constants vs magic numbers, but the code review needs to be an all-or-nothing affair. If every single thing is not attended to, then it casts doubt around the entire body of work.

PR #8, opened 2016-04-15
Implemented ldap merge poller

theunissens:
Implemented ldap merge poller, merging ldap users into imqs for authorization when ldap delegated for authentication

### DIFF
--- a/all_test.go
+++ b/all_test.go
@@ -26,12 +26,14 @@ Suggested test runs that you should do:
 	go test github.com/IMQS/authaus -test.cpu 2 -backend_postgres
 	go test -race github.com/IMQS/authaus -test.cpu 2
 	go test -race github.com/IMQS/authaus -test.cpu 2 -backend_postgres
+	go test -race github.com/IMQS/authaus -test.cpu 2 -backend_postgres -backend_ldap
 
 In other words, test with and without race detector, and with dummy backend,
 and Postgres backend. I'm not sure that testing without the race detector adds any value.
 */
 
 var backend_postgres = flag.Bool("backend_postgres", false, "Run tests against Postgres backend")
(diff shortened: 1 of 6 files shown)


all_test.go
```diff
@@ -173,112 +189,129 @@ func isPrefix(prefix, str string) bool {
...
+		}
 
-	teardown(c)
+		Teardown(c)
+	}
 }
 
 func TestRenameIdentity(t *testing.T) {
-	c := setup1(t)
+	if !*backend_ldap {
```
bmharper: Rather just make these functions exit early, if *backend_ldap is true.
Same applies to TestResetPassword, and I'm guessing in other places too.

bmharper: And use `defer Teardown()` at the top of the function.

auth.go
```diff
@@ -214,6 +214,7 @@ type Central struct {
 	Log                    *log.Logger
 	MaxActiveSessions      int32
 	NewSessionExpiresAfter time.Duration
+	Ticker *time.Ticker
```
bmharper: It looks like this code hasn't been formatted with gofmt

ldap.go
```diff
@@ -15,87 +18,459 @@ const (
...
 }
 
 func (x *ldapUserStore) RenameIdentity(oldIdent, newIdent string) error {
 	return ErrUnsupported
 }
 
-func (x *ldapUserStore) GetIdentities() ([]string, error) {
-	return []string{}, ErrUnsupported
+func (x *ldapUserStore) GetIdentities() ([]AuthUser, error) {
+	rows, err := x.db.Query(`SELECT userid, email, username, firstname, lastname, mobile FROM authuserstore WHERE (archived = false OR archived IS NULL) ORDER BY username`)
```
bmharper: This is a mega-nasty abstraction failure, but we can deal with all of the cleanup after the demo.

PR #9, opened 2016-05-09
Restructured LDAP and Authentication implementations

theunissens:


auth.go
```diff
@@ -40,6 +42,7 @@ var (
 	ErrInvalidSessionToken    = errors.New("Invalid session token")
 	ErrInvalidPasswordToken   = errors.New("Invalid password token")
 	ErrPasswordTokenExpired   = errors.New("Password token has expired")
+	ldapUsed                  = false
```
bmharper: A global variable like this (ldapUsed) doesn't belong here. Classic scope leakage.

auth.go
```diff
@@ -24,6 +24,8 @@ const (
 	sessionTokenLength = 30
 
 	defaultSessionExpirySeconds = 30 * 24 * 3600
+
+	tickerTime = 5
```
bmharper: Time constants must either have a unit in their name, such as "tickerTimeSeconds", or (preferably) be strongly typed, such as 
```go
const tickerTime = 5 * time.Second
```
Also, this variable needs a better name. Nothing tells us that it is the ticker interval for the LDAP identities merge.

auth.go
```diff
@@ -219,14 +222,21 @@ type Central struct {
...
-func NewCentral(logfile string, authenticator Authenticator, userStore UserStore, permitDB PermitDB, sessionDB SessionDB, roleGroupDB RoleGroupDB) *Central {
+func NewCentral(logfile string, ldap LDAP, userStore UserStore, permitDB PermitDB, sessionDB SessionDB, roleGroupDB RoleGroupDB) *Central {
 	c := &Central{}
-	c.authenticator = &sanitizingAuthenticator{
-		backend: authenticator,
+	if ldap != nil {
+		c.ldap = &sanitizingLDAP{
+			backend: ldap,
+		}
+		ldapUsed = true
```
bmharper: Still need to get rid of this global `ldapUsed`

auth.go
```diff
@@ -279,12 +289,11 @@ func NewCentralFromConfig(config *Config) (central *Central, err error) {
 		panic(errors.New("SessionExpirySeconds must be 0 or more"))
 	}
 
-	//var ldapAuth LdapAuthenticator
-	if auth, err = createAuthenticator(&config.Authenticator); err != nil {
-		panic(err)
+	if ldap, err = NewAuthenticator_LDAP(&config.LDAP); err != nil {
+		panic(errors.New(fmt.Sprintf("Error creating LDAP Authenticator: %v", err)))
```
bmharper: Can use fmt.Errorf here

auth.go
```diff
@@ -474,32 +435,87 @@ func (x *Central) Login(identity, password string) (sessionkey string, token *To
 }
 
 // Merges ldap with user store every 10 seconds
-func (x *Central) Merge() error {
+func (x *Central) MergeTicker() error {
 	x.Log.Info("Starting LDAP merge process")
-	x.Ticker = time.NewTicker(time.Second * 10)
+	x.Ticker = time.NewTicker(ldapMergeTickerSeconds)
```
bmharper: This function should be named "StartMergeTicker" so make it clear that it is launching a process that is going to run for some time. In our case, indefinitely.

The control flow of this function is more complicated than it needs to be. You can just do
```go
go func() {
    x.Ticker := time.NewTicker(ldapMergeTickerSeconds)
    for _ = range x.Ticker.C {
        // do stuff
    }
}()
```

I don't think we should emit log messages stating that we are about to do a merge. Only emit log messages when something changes, or when an error occurs. Otherwise our log files end up with a million "starting merge/finishing merge" messages, which don't add any signal to the noise.
It is however useful to have timing information about the merge process. What I would do is to have a counter that increments every time a merge occurs. Only if that counter mod 60 equals zero, then emit the time. That way, we get one timing message per hour, which is reasonable. Also, count the time duration in the program, and emit that directly into the logs (instead of emitting two absolute time measurements).

I don't think that the ticker should abort if it encounters an error. If it encounters an error, it should log that failure, and then proceed as usual. Think about what happens if we abort on error. It means that the merge ticker kills itself, and so DevOps has to logon to that server and restart the auth service, which is definitely not what we want.

auth.go
```diff
@@ -474,32 +435,87 @@ func (x *Central) Login(identity, password string) (sessionkey string, token *To
...
+
 		}
-		x.Ticker.Stop()
 	}()
 
 	return nil
+}
+
+// Merges the ldap users and imqs users
+func (x *Central) Merge(ldapUsers []AuthUser, imqsUsers []AuthUser) error {
```
bmharper: This function needs a more descriptive name, such as "MergeIntoLocalUserStore"

auth.go
```diff
@@ -474,32 +435,87 @@ func (x *Central) Login(identity, password string) (sessionkey string, token *To
...
+	found := false
```
bmharper: `found` does not need to exist outside the scope of the `range ldapUsers` loop, and therefore it MUST be inside that loop. Having `found` be in the outer scope make it significantly harder to understand the code. Same applies to the Archive case.

I think it's important that we do a check here to avoid needless calls to UpdateIdentity. This allows us to log whenever a meaningful Update occurs. You'll want to create an "equals" function on `AuthUser` for this.

auth.go
```diff
@@ -474,32 +435,87 @@ func (x *Central) Login(identity, password string) (sessionkey string, token *To
...
+	// Insert or update
+	for _, ldapUser := range ldapUsers {
+		for _, imqsUser := range imqsUsers {
+			if ldapUser.Username == imqsUser.Username {
```
bmharper: You need to create maps here, where the keys are the usernames, to avoid the N^2 execution time. If a client has 1000 LDAP users (a very reasonable expectation), then we have 2 million string comparisons to do. After implementing this change, it will be even clearer why `found` needs to be more tightly scoped, because you'll write a single line:
```go
imqsUser, found := usernameToImqs[ldapUser.Username]
```

auth.go
```diff
@@ -678,7 +694,6 @@ func (x *Central) Close() {
 	}
 	if x.Ticker != nil {
 		x.Ticker.Stop()
-		x.Ticker = nil
```
bmharper: Setting`x.Ticker = nil` is the right thing to do, and should work correctly with the simpler code flow that I mentioned in an earlier comment

config.go
```diff
@@ -24,11 +24,13 @@ Full populated config:
 		"Port":			8080,
 		"Bind":			"127.0.0.1"
 	},
-	"Authenticator": {
-		"Type":			"ldap",
-		"LdapHost":		"domaincontroller.example.com",
+	"LDAP": {
```
bmharper: Rather keep "imqs" out of the examples. Use example.com and other names like it.

db.go
```diff
@@ -20,32 +20,36 @@ var (
...
 	GetUserIdFromIdentity(identity string) (UserId, error)                                              // Gets the userid from the identity supplied
 	GetIdentities() ([]AuthUser, error)                                                                 // Retrieve a list of all identities
-	GetLdapUsers() ([]AuthUser, error)                                                                  // Retrieve the list of users from ldap
-	Merge([]AuthUser) error                                                                             // Merges ldap users with our userstore
 	Close()                                                                                             // Typically used to close a database handle
 }
 
+// The LDAP interface allows authentication and the ability to retrive the LDAP's users and merge them into our system
+type LDAP interface {
+	Authenticate(identity, password string) error // Return nil error if the password is correct, otherwise one of ErrIdentityAuthNotFound or ErrInvalidPassword
```
bmharper: Nitpicky: No need to say "Return nil error". Just say "Return nil"

db.go
```diff
@@ -90,13 +94,12 @@ type LdapUser struct {
...
 // Authenticator/Userstore that simply stores identity/passwords in memory
-type dummyUserStoreAndAuth struct {
+type dummyUserStore struct {
 	users     map[UserId]*dummyUser
 	usersLock sync.RWMutex
 }
 
-type dummyLdapUserStoreAndAuth struct {
-	imqsUsers map[UserId]*dummyUser
+type dummyLdap struct {
```
bmharper: The LdapUser type is only ever used by dummyLdap, so it should rather be named as such, and not exported from authaus.

db.go
```diff
@@ -451,52 +333,63 @@ func (x *dummyLdapUserStoreAndAuth) GetLdapUsers() ([]AuthUser, error) {
...
-type sanitizingAuthenticator struct {
-	backend Authenticator
-}
-
 // UserStore that sanitizes inputs, so that we have more consistency with different backends
 type sanitizingUserStore struct {
 	backend UserStore
 }
 
+// UserStore that sanitizes inputs, so that we have more consistency with different backends
```
bmharper: Comments are copy + pasted, without being altered.

db_sql.go
```diff
@@ -159,12 +148,23 @@ func (x *sqlUserStoreDB) CreateIdentity(email, username, firstname, lastname, mo
 	// Insert into user store
 	if tx, etx := x.db.Begin(); etx == nil {
 		// Check if the user exists (is not archived)
-		row := tx.QueryRow("SELECT email FROM authuserstore WHERE LOWER(email) = $1 AND archived = false", CanonicalizeIdentity(email))
-		scanErr := row.Scan()
-		if strings.Index(scanErr.Error(), "no rows in result set") == -1 {
-			tx.Rollback()
-			scanErr = ErrIdentityExists
-			return NullUserId, scanErr
+		if ldapUsed {
```
bmharper: Whether LDAP is sitting in front of this or not, does not belong here. That decision needs to happen before we reach sqlUserStoreDB. sqlUserStoreDB should be concerned only with it's own role. The word LDAP shouldn't appear in the methods of sqlUserStoreDB, except maybe inside a comment.
You might want to make these checks for user existence little standalone functions, to make the code more readable.

db_sql.go
```diff
@@ -174,20 +174,22 @@ func (x *sqlUserStoreDB) CreateIdentity(email, username, firstname, lastname, mo
...
 			tx.Rollback()
 			return NullUserId, scanErr
 		}
 
-		// Insert into auth user
-		if _, eCreateAuthUser := tx.Exec(`INSERT INTO authuserpwd (userid, password) VALUES ($1, $2)`, userId, hash); eCreateAuthUser != nil {
-			//fmt.Printf("Insert into authuserpwd failed because: %v\n", eCreateAuthUser)
-			if strings.Index(eCreateAuthUser.Error(), "duplicate key") != -1 {
-				eCreateAuthUser = ErrIdentityExists
+		if !ldapUsed {
```
bmharper: Instead of referring to a global flag here, sqlUserStoreDB could use a member variable called 'enableAuthenticator'. Either way - no globals like that.

db_sql.go
```diff
@@ -300,7 +302,7 @@ func (x *sqlUserStoreDB) GetIdentityFromUserId(userId UserId) (string, error) {
 }
 
 func (x *sqlUserStoreDB) GetUserIdFromIdentity(identity string) (UserId, error) {
-	row := x.db.QueryRow(`SELECT userid FROM authuserstore WHERE LOWER(email) = $1 AND (archived = false OR archived IS NULL)`, CanonicalizeIdentity(identity))
+	row := x.db.QueryRow(`SELECT userid FROM authuserstore WHERE LOWER(email) = $1 OR LOWER(username) = $1 AND (archived = false OR archived IS NULL)`, CanonicalizeIdentity(identity))
```
bmharper: Before committing this work, we need to document exactly what "identity" means... IIRC in our last discussion Identity was going to mean "Either a username or an email address. If ambiguous, the system must try both".

ldap.go
```diff
@@ -18,146 +17,39 @@ const (
...
-		if err := rows.Scan(&user.userid, &user.email, &user.username, &user.firstname, &user.lastname, &user.mobilenumber); err != nil {
-			return []AuthUser{}, err
-		}
-		result = append(result, AuthUser{UserId(user.userid.Int64), user.email.String, user.username.String, user.firstname.String, user.lastname.String, user.mobilenumber.String})
+		return err
 	}
-	if rows.Err() != nil {
-		return []AuthUser{}, rows.Err()
+	err = con.Bind(fmt.Sprintf(`%v@%v`, identity, x.config.LdapDomain), password)
+	if err != nil {
```
bmharper: It's not good to just blanket all errors as "Identity not found". At the very least, we need to discriminate between the following 3 cases:

1. Identity not found
2. Password incorrect
3. Other (with error string)

Also, functions returning a single `error` should not used named return values.

ldap.go
```diff
@@ -166,14 +58,19 @@ func (x *ldapUserStore) GetLdapUsers() ([]AuthUser, error) {
...
 		"dc=imqs,dc=local",
 		ldap.ScopeWholeSubtree, ldap.DerefAlways, 0, 0, false,
 		"(objectclass=user)",
 		attributes,
 		nil)
 
-	sr, err := x.con.Search(search_request)
+	con, err := NewLDAPConnectAndBind(x.config)
+	defer con.Close()
+	if err != nil {
```
bmharper: I'm pretty sure you need to do the error checking before the `con.Close()`, because if the bind fails, then `con` will most likely be null, in which case you'll end up with a panic. There needs to be a unit test for this case.

all_test.go
```diff
@@ -35,13 +39,29 @@ and Postgres backend. I'm not sure that testing without the race detector adds a
 var backend_postgres = flag.Bool("backend_postgres", false, "Run tests against Postgres backend")
 var backend_ldap = flag.Bool("backend_ldap", false, "Run tests against LDAP backend")
 
+// These are hard-coded identities for unit test prediction
```
bmharper: "unit test predictability"

all_test.go
```diff
@@ -53,19 +73,28 @@ var conx_postgres = DBConnection{
 	SSL:      false,
 }
 
+var conx_ldap = ConfigLDAP{
+	Encryption:   "",
+	LdapDomain:   "imqs.local",
+	LdapHost:     "imqs.local",
+	LdapPassword: joePwd,
+	LdapUsername: fmt.Sprintf(joeIdentity, ".imqs.local"),
```
bmharper: This statement doesn't make any sense.

all_test.go
```diff
@@ -115,42 +142,60 @@ func setup1(t *testing.T) *Central {
...
 	if *backend_ldap {
-		err := central.Merge()
+		if !*backend_postgres {
+			ldapTest.AddLdapUser(joeIdentity, joePwd, "joe@gmail.com", "Firstname", "Lastname", "Mobilenumber")
+			ldapTest.AddLdapUser(jackIdentity, jackPwd, "jack@gmail.com", "Firstname", "Lastname", "Mobilenumber")
+			ldapTest.AddLdapUser(SamIdentity, SamPwd, "Sam@gmail.com", "Firstname", "Lastname", "Mobilenumber")
+			ldapTest.AddLdapUser(iHaveNoPermitIdentity, iHaveNoPermitPwd, "iHaveNoPermit@gmail.com", "Firstname", "Lastname", "Mobilenumber")
+		}
+
+		err := central.StartMergeTicker()
```
bmharper: I don't think it's a good idea to run the merge ticker during most unit tests. It will make tests much less predictable. I would run the merge ticker only during tests that explicitly test the merge ticker (eg TestMergeLdap). Inside that test, you'll want to have the merge ticker running on a much tighter frequency. I would suggest 10000 times a second. Then, stress parts of the system that are likely to be in race conditions with the merge ticker. For example the functions Login and GetAuthenticatorIdentities. This is the kind of test that needs to be run with the race detector turned on. You'll want to run that kind of loop exhaustively for a few seconds (say 10 seconds), all the time doing as many Login and GetAuthenticatorIdentities as the machine can do.

all_test.go
```diff
@@ -368,50 +447,50 @@ func BenchmarkScrypt256(b *testing.B) {
...
-		go doLogin(i, conPerThread, waits[i])
-	}
-	for i := 0; i < nsimul; i++ {
-		_, ok := <-waits[i]
-		if !ok {
-			t.Errorf("Channel closed prematurely")
-		}
-	}
-}
+//// This test must be run with at least 2 processors "go test -test.cpu 2"
```
bmharper: Why is this test commented out?

auth.go
```diff
@@ -24,6 +24,8 @@ const (
 	sessionTokenLength = 30
 
 	defaultSessionExpirySeconds = 30 * 24 * 3600
+
+	ldapMergeTickerSeconds = 5 * time.Second
```
bmharper: This seems way too frequent. I would imagine every minute is more reasonable. However, for the sake of unit tests, I'm guessing this needs to be configurable. Once again, don't make it a global variable. Make the default a global variable, but make it a configurable member of Central.

auth.go
```diff
@@ -215,18 +218,25 @@ type Central struct {
 	MaxActiveSessions      int32
 	NewSessionExpiresAfter time.Duration
 	Ticker                 *time.Ticker
```
bmharper: I don't think Ticker needs to be public. Also, it needs a better name, like userStoreMergeTicker

all_test.go
```diff
@@ -189,16 +234,45 @@ func isPrefix(prefix, str string) bool {
...
+			if _, _, err := c.Login(joeIdentity, joePwd); err != nil {
+				t.Fatalf("Login should have successed, but error was : %v", err)
+			}
+		} else {
+			// Add new users to ldap, wait for merge then login
+			ldapTest.AddLdapUser(tomIdentity, tomPwd, "tomh@hotgmail.com", "tom", "hanks", "08467531243")
+			tomhPermit := setup1_permit()
+			c.permitDB.SetPermit(tomUserId, &tomhPermit)
+			//Now we wait for a merge to take place
+			time.Sleep(ldapMergeTickerSeconds*2)
```
bmharper: I don't think we need to test whether the merge ticker actually runs. That code is so predictable that it's almost impossible to get wrong. I mean - the part where we do `for _ := range ticker ...`. And that is actually the thing that's being tested by waiting for the ticker to work. Instead, I would just not run the ticker during tests like this, and instead call the merge function manually. You don't want a unit test like this to take 25 seconds, when the real functionality that's being tested, can be run in under 1 second.
I would keep this test as is, except remove all the waiting. Then, add another test which stresses the possible races, as I described in an earlier comment.

all_test.go
```diff
@@ -24,8 +25,11 @@ Suggested test runs that you should do:
 
 	go test github.com/IMQS/authaus -test.cpu 2
 	go test github.com/IMQS/authaus -test.cpu 2 -backend_postgres
+	go test github.com/IMQS/authaus -test.cpu 2 -backend_ldap
+	go test github.com/IMQS/authaus -test.cpu 2 -backend_ldap -backend_postgres
 	go test -race github.com/IMQS/authaus -test.cpu 2
 	go test -race github.com/IMQS/authaus -test.cpu 2 -backend_postgres
+	go test -race github.com/IMQS/authaus -test.cpu 2 -backend_ldap
```
bmharper: I think there should be some comments describing what happens under the various combinations of backend_postgres, backend_ldap. ie a matrix of what the test setup looks like with each of those flags on and off.

bmharper:
I'm happy that the architecture looks better now. It would be great to also report some code coverage numbers for the authaus package. It's the kind of thing that deserves close to 100% line coverage.

ldap.go
navonod: If we're sending a specific error to be displayed on the front end, is it not viewed as a security risk? E.g.: Allowing any potential malicious user to know that a password failed on an existing user gives him sufficient knowledge to try a brute force login for that user. Yes, we can have other measures to prevent brute-forcing, but isn't this the first line of defence? A single, blanketed error on the front-end is kind of industry-standard security these days.

ldap.go
```diff
@@ -18,146 +17,37 @@ const (
...
-			return []AuthUser{}, err
-		}
-		result = append(result, AuthUser{UserId(user.userid.Int64), user.email.String, user.username.String, user.firstname.String, user.lastname.String, user.mobilenumber.String})
+		return err
 	}
-	if rows.Err() != nil {
-		return []AuthUser{}, rows.Err()
+	err = con.Bind(fmt.Sprintf(`%v@%v`, identity, x.config.LdapDomain), password)
+	if err != nil {
+		er = NewError(ErrInvalidCredentials, err.Error())
```
navonod: Assuming that we care about distinguishing between types of auth errors, shouldn't we be checking the error here, before assuming it to be invalid credentials?

PR #10, opened 2016-06-29
Added authusertype to userobject and to authuserstore table

theunissens:


auth.go
```diff
@@ -210,12 +210,11 @@ type Central struct {
 	// on their operands.
 	Stats                  CentralStats
 	ldap                   LDAP
-	authenticator          Authenticator
 	userStore              UserStore
 	permitDB               PermitDB
 	sessionDB              SessionDB
 	roleGroupDB            RoleGroupDB
-	renameLock             sync.Mutex
+	mutex                  sync.Mutex
```
bmharper: The first rule of locks is that it needs to be *extremely* clear what the lock governs. Imagine a house covered in locks, but you don't know which lock opens which door. That's what it's like when you see a variable named "mutex". Unless it's painfully obvious, every mutex/lock need to have a comment next to it, explaining what it guards.

auth.go
```diff
@@ -385,7 +380,24 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
...
-	if eAuth := x.authenticator.Authenticate(identity, password); eAuth == nil {
+	user, eGetUser := x.userStore.GetUser(userId)
+	if eGetUser != nil {
+		return nil, ErrIdentityAuthNotFound
+	}
+	var eAuth error
+	if user.Type == LDAPUserType {
+		eAuth := x.ldap.Authenticate(identity, password)
+		// We want to return Invalid Password or IdentityAuthNotFound, not Invalid Credentials
+		// as LDAP doesnt differentiate between the 2
```
bmharper: "doesn't differentiate between the two"

auth.go
```diff
@@ -410,16 +422,43 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
...
+	}
+
+	token = &Token{}
+	if user.Type == LDAPUserType {
+		err = x.ldap.Authenticate(identity, password)
+		// We want to return Invalid Password or IdentityAuthNotFound, not Invalid Credentials
+		// as LDAP doesnt differentiate between the 2
+		if err == ErrInvalidCredentials {
+			if identity != user.Username {
+				// If the user logged in with his email address, we would find a userid,
```
bmharper: This sentence isn't valid english

db.go
```diff
@@ -152,7 +157,7 @@ func (x *dummyUserStore) Close() {
 func (x *dummyUserStore) SetPassword(userId UserId, password string) error {
 	x.usersLock.Lock()
 	defer x.usersLock.Unlock()
-	if user, exists := x.users[userId]; exists {
+	if user, exists := x.users[userId]; exists && user.authUserType == IMQSUserType {
```
bmharper: Rather create a method `(u AuthUserType) CanSetPassword() bool` and use that method here and wherever else the logic applies.

db_sql.go
```diff
@@ -179,7 +177,7 @@ func (x *sqlUserStoreDB) CreateIdentity(email, username, firstname, lastname, mo
 			}
 		}
 
-		if x.enableAuthenticator {
+		if authUserType == IMQSUserType {
 			// Insert into auth user
```
bmharper: This comment is a lie - there is no "auth user" table - just remove the comment.

db.go
```diff
@@ -8,6 +8,8 @@ import (
 )
 
 const (
+	IMQSUserType AuthUserType = 0
```
bmharper: Place enum-like constants in their own "const" block. So AuthUserType constants go in their own block, and NullUserId in it's own block, etc. If it's just one constant, then it doesn't need a const block.

Find a different name for IMQSUserType, which doesn't use the word "IMQS". Until now, authaus hasn't had any explicit tie to IMQS, and it is good to keep it that way. So rather call it "BuiltinUser", or something like that.
Start the name with UserType, such as UserTypeLDAP, UserTypeBuiltin.

Since these constants are embedded in the database, you need to make that very clear. There must be comments where the constants (enums) are declared, and there needs to be an accompanying comment in the DB migration code explaining that the field takes values from the AuthUserType constants.

Do we really need to call it AuthUserType? Can it not just be UserType?

bmharper:
There is zero documentation around these changes. All of this new functionality needs to be explained inside the main authaus docs. And the english needs to be excellent!

auth.go
```diff
@@ -410,16 +422,43 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
...
+				// but the authentication would fail, we need to check if he logged in with his
+				// username, if he did, then we can just return invalid password, else we return
+				// identity not found
+				err = ErrIdentityAuthNotFound
```
bmharper: This logic doesn't make sense:
1. We successfully found a user in our system for this "identity"
2. We claim that this identity is not found
Something is wrong with this picture.

I don't see the point in disallowing a login because the user entered his email address instead of his username.

auth.go
```diff
@@ -385,7 +380,24 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
 	if eGetIdent != nil {
 		return nil, ErrIdentityAuthNotFound
 	}
-	if eAuth := x.authenticator.Authenticate(identity, password); eAuth == nil {
+	user, eGetUser := x.userStore.GetUser(userId)
+	if eGetUser != nil {
+		return nil, ErrIdentityAuthNotFound
+	}
+	var eAuth error
+	if user.Type == LDAPUserType {
```
bmharper: Why are we calling Authenticate(identity, password), when we always go to the trouble of first fetching the UserID. Surely we should make a function called Authenticate(userID, password). That would remove the need for all callers of Authenticate to code this special logic about turning ErrInvalidCredentials into ErrInvalidPassword.

auth.go
```diff
@@ -210,12 +210,11 @@ type Central struct {
 	// on their operands.
 	Stats                  CentralStats
 	ldap                   LDAP
-	authenticator          Authenticator
 	userStore              UserStore
 	permitDB               PermitDB
 	sessionDB              SessionDB
 	roleGroupDB            RoleGroupDB
-	renameLock             sync.Mutex
+	renameLock                  sync.Mutex //
```
bmharper: gofmt?

auth.go
```diff
@@ -410,18 +402,13 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
 // The internal session expiry is controlled with the member NewSessionExpiresAfter.
 // The session key is typically sent to the client as a cookie.
 func (x *Central) Login(identity, password string) (sessionkey string, token *Token, err error) {
-	token = &Token{}
-	if err = x.authenticator.Authenticate(identity, password); err != nil {
+	userId, authErr := x.authenticate(identity, password)
```
bmharper: phew, much better

auth.go
```diff
@@ -667,26 +663,26 @@ func (x *Central) ArchiveIdentity(userId UserId) error {
 	return nil
 }
 
-// Get userid from identity.
-func (x *Central) GetUserIdFromIdentity(identity string) (UserId, error) {
-	userId, e := x.userStore.GetUserIdFromIdentity(identity)
+// Get AuthUser object from identity.
+func (x *Central) GetUserForIdentity(identity string) (AuthUser, error) {
```
bmharper: The use of the word "For" here is inconsistent, and generally doesn't comprehend very well. Rather use "From". Same for GetUserForUserId. Eg `GetUserFromIdentity`, `GetUserFromId`

db.go
```diff
@@ -13,42 +13,83 @@ const (
...
+	}
+}
+
+func (u AuthUserType) CanRenameIdentity() bool {
+	switch u {
+	case UserTypeDefault:
+		return true
+	default:
+		return false
+	}
```
bmharper: nice. much cleaner

db.go
```diff
@@ -13,42 +13,83 @@ const (
 	NullUserId              UserId = 0
 )
 
+// The DB relies on these constants to be these exact value. If their values are changed, things can go pear-shaped.
```
bmharper: Comments like these should rather be precise. You don't need to say that things will go pear-shaped. Rather let the reader make that decision for himself. For example:
`These constants are embedded inside our database (in the table AuthUserStore). They may never change.`

db_sql.go
```diff
@@ -42,8 +42,7 @@ const (
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 type sqlUserStoreDB struct {
-	enableAuthenticator bool
-	db                  *sql.DB
+	db *sql.DB
```
bmharper: seriously - enable gofmt so that it runs every time you hit "save". It REALLY is the right way to setup your dev environment. I have tried the other options. That really is best.

PR #11, opened 2016-07-28
Bug fixes: TG-633 and TG-641

theunissens:


auth.go
```diff
@@ -355,7 +355,7 @@ func (x *Central) GetTokenFromSession(sessionkey string) (*Token, error) {
 		x.Stats.IncrementInvalidSessionKey(x.Log)
 		return token, err
 	} else {
-		if time.Now().UnixNano() > token.Expires.UnixNano() {
+		if time.Now().UTC().UnixNano() > token.Expires.UnixNano() {
```
bmharper: Look at the implementation of time.UTC and time.UnixNano, and you'll see that adding "UTC()" in here has no effect, so don't do it.

db_sql.go
```diff
@@ -304,14 +304,22 @@ func (x *sqlUserStoreDB) GetIdentities() ([]AuthUser, error) {
...
 }
 
 func (x *sqlUserStoreDB) GetUserFromUserId(userId UserId) (AuthUser, error) {
-	return x.getUser(x.db.QueryRow("SELECT userid, email, username, firstname, lastname, mobile, authusertype FROM authuserstore WHERE userid = $1 AND (archived = false OR archived IS NULL)", userId))
+	return getUserFromUserId(x.db, userId)
 }
 
-func (x *sqlUserStoreDB) getUser(row *sql.Row) (AuthUser, error) {
+func getUserFromIdentity(db *sql.DB, identity string) (AuthUser, error) {
+	return getUser(db.QueryRow("SELECT userid, email, username, firstname, lastname, mobile, authusertype FROM authuserstore WHERE (LOWER(email) = $1 AND (archived = false OR archived IS NULL)) OR (LOWER(username) = $1 AND (archived = false OR archived IS NULL))", CanonicalizeIdentity(identity)))
```
bmharper: The "ARCHIVED" part of the logic can be hoisted out so that it is only expressed once.
`(a AND x) OR (b AND x) === (a OR b) AND x`

bmharper:
Can you explain EXACTLY what is happening with dates and the Postgres DB?
In other words:
1. What happens if you insert a SAST time into a TIMESTAMP column, and what happens if you insert a UTC time?
2. What happens when you read a time out of the DB? Does it come back with "Location == UTC"?

If my guess in (2) is correct, then you ONLY need to use time.UTC() the moment before inserting a time value into the DB. And if this is so, then don't litter the code with any other calls to UTC(), because doing so implies it's necessary, and that is going to cause a lot of confusion.

all_test.go
```diff
@@ -720,6 +733,34 @@ func TestMaxSessionLimit(t *testing.T) {
 	}
 }
 
+func TestSqlSessionCache(t *testing.T) {
+	c := setup1(t)
+	defer Teardown(c)
+	c.MaxActiveSessions = 1
+
+	// Login first time
+	key, _, _ := c.Login(joeIdentity, joePwd)
```
bmharper: Is this intended to test the capitalization of the login name? If so, we should say so in comments, and also login with a differently cased username to the actual "joeIdentity"

auth.go
```diff
@@ -440,10 +440,10 @@ func (x *Central) Login(identity, password string) (sessionkey string, token *To
 	return sessionkey, token, nil
 }
 
-func (x *Central) authenticate(identity, password string) (UserId, error) {
+func (x *Central) authenticate(identity, password string) (UserId, string, error) {
```
bmharper: UserId is obvious, and error is obvious, but what is the "string" that is returned here? It should be specified in a comment or with named return values (I'd go for comment in this case).

all_test.go
theunissens: It is not. We login here to generate a session key and remove it from the memory. Then we test the session cache, which would then hit the db. I should probably change the name of the test to: "TestDBSessionCache"

PR #15, opened 2016-09-26
TG-702: Refactored tests, splitting ldap integrated tests from auth t…

theunissens:
…ests

### DIFF
--- a/all_test.go
+++ b/all_test.go
@@ -22,35 +22,15 @@ Create a test Postgres database:
 
 Suggested test runs that you should do:
 
-	---- Without race detector ---
-
 	-- Test using maps/arrays mocking the backend:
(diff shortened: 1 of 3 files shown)


all_test.go
```diff
@@ -25,32 +25,26 @@ Suggested test runs that you should do:
 	---- Without race detector ---
 
 	-- Test using maps/arrays mocking the backend:
-	go test github.com/IMQS/authaus -test.cpu 2
+	go test github.com/IMQS/authaus -test.cpu 2 -run TestAuth
```
bmharper: Generally, looks great.
This comment is a bit too long though. Rather don't duplicate all of the instructions with and without "-race". The user is smart enough to figure that out. Just leave the "-race" in all of the examples.

PR #16, opened 2016-12-05
TG-709: Changed login authentication to check on both email and usern…

navonod:
…ame fields

### DIFF
--- a/db.go
+++ b/db.go
@@ -16,7 +16,7 @@ const (
 // These constants are embedded inside our database (in the table AuthUserStore). They may never change.
 const (
 	UserTypeDefault AuthUserType = 0
-	UserTypeLDAP                 = 1
+	UserTypeLDAP    AuthUserType = 1
 )
 
 var (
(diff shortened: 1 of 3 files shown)


db.go
```diff
@@ -68,7 +68,7 @@ func (u AuthUserType) UsernameAsIdentity() bool {
 // It is also responsible for adding new users, changing passwords etc.
 // All operations except for Close must be thread-safe.
 type UserStore interface {
-	Authenticate(identity, password string) error                                                                                  // Return nil error if the username and password are correct, otherwise one of ErrIdentityAuthNotFound or ErrInvalidPassword
+	Authenticate(user string, password string) error                                                                               // Return nil error if the username and password are correct, otherwise one of ErrIdentityAuthNotFound or ErrInvalidPassword
```
theunissens: Not necessary to put the string there.

theunissens: Also, why are we changing it from identity to user? I think identity is perfectly fine. We also kind of refer to the username/email as identity throughout the code base, so changing the interface would mean it would be inconsistent.

PR #18, opened 2017-01-23
Added a getIdentity() method to an AuthUser, to determine which field (of email or us…

navonod:
…ername) to return to request queries. This is viable because we now authenticate against either the username or the email field. We define the identity as either one of these two fields.

### DIFF
--- a/db.go
+++ b/db.go
@@ -853,3 +853,10 @@ func (x *dummyPermitDB) SetPermit(userId UserId, permit *Permit) error {
 
 func (x *dummyPermitDB) Close() {
 }
+
+func (a *AuthUser) getIdentity() string {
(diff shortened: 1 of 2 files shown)


db.go
```diff
@@ -853,3 +853,10 @@ func (x *dummyPermitDB) SetPermit(userId UserId, permit *Permit) error {
 
 func (x *dummyPermitDB) Close() {
 }
+
+func (a *AuthUser) getIdentity() string {
+	if (&a.Email == nil || len(a.Email) == 0) {
```
bmharper: Why do you check if the address of a.Email is nil?

navonod: I've seen it done this way in other go code - is it safe to assume the address will never be nil?

PR #19, opened 2017-02-15
PCS-1286 - LDAP Backwords Compatibilty

theunissens:
Fixed an ldap error, where if you remove the ldap config, or deactivate
ldap, a nullpointer exception would occur if you try and login with an
ldap user.


bmharper:
Looks good

PR #23, opened 2017-08-14
TI-2181: User profile enhancements

dbferreira:


db.go
```diff
@@ -50,13 +50,13 @@ func (u AuthUserType) CanRenameIdentity() bool {
...
+	ResetPasswordStart(userId UserId, expires time.Time) (string, error)                                                                                                                                                         // Create a one-time token that can be used to reset the password with a subsequent call to ResetPasswordFinish
+	ResetPasswordFinish(userId UserId, token string, password string) error                                                                                                                                                      // Check that token matches the last one generated by ResetPasswordStart, and if so, call SetPassword
+	CreateIdentity(email, username, firstname, lastname, mobilenumber, telephonenumber, remarks string, created int64, createdby string, modified int64, modifiedby, password string, authUserType AuthUserType) (UserId, error) // Create a new identity. If the identity already exists, then this must return ErrIdentityExists.
```
bmharper: I think we should separate the non-essential fields out into a "OptionalUserFields" struct. These would include:
* firstname
* lastname
* mobilenumber
* telephonenumber
* remarks
* created
* createdby
* modified
* modifiedby

Or, perhaps, just take an `AuthUser` struct, and do reasonable things like ignore UserId when creating a new identity.

bmharper: `created` and `modified` should be `time.Time`

db_sql.go
```diff
@@ -780,6 +792,15 @@ func createMigrations() []migration.Migrator {
 
 		// 8. We add AuthUserType field to the userstore, to determine what type of user account this is.
 		`ALTER TABLE authuserstore ADD COLUMN authusertype SMALLINT default 0;`,
+
+		// 9. Additional data fields as well as fields to keep track of changes to users
+		`ALTER TABLE authuserstore 
+			ADD COLUMN phone VARCHAR, 
+			ADD COLUMN remarks VARCHAR,
+			ADD COLUMN created BIGINT,
```
bmharper: I think rather use TIMESTAMP fields for time. The only thing you need to be careful about then, is to make sure you convert all times to UTC before inserting them into the DB.

db.go
```diff
@@ -464,7 +480,7 @@ func (x *sanitizingUserStore) ResetPasswordFinish(userId UserId, token string, p
 	return x.backend.ResetPasswordFinish(userId, token, password)
 }
 
-func (x *sanitizingUserStore) CreateIdentity(email, username, firstname, lastname, mobilenumber, password string, authUserType AuthUserType) (UserId, error) {
+func (x *sanitizingUserStore) CreateIdentity(email, username, firstname, lastname, mobilenumber, telephonenumber, remarks string, created int64, createdby string, modified int64, modifiedby, password string, authUserType AuthUserType) (UserId, error) {
```
bmharper: I think all of these function signatures would benefit from passing a struct around, instead the long list of arguments that this has become.

PR #24, opened 2017-08-16
TI-2093 - Audit user login actions

stewartm83:
Please review.


auth.go
```diff
@@ -444,6 +452,39 @@ func (x *Central) Login(identity, password string) (sessionkey string, token *To
 	return sessionkey, token, nil
 }
 
+func auditUserAction(identity, password, actionDescription string) error {
+	auditServiceUrl := "http://localhost:2016/auth-logaction"
```
theunissens: This url needs to be configured

theunissens: I feel like auth.go is becoming quite big, perhaps we should create another file audit.go where this function can live?

theunissens:
I would mock out the audit service using httptest.NewServer and just write a unit test for this new functionality. There are examples of this done in all_test.go in imqsauth project, the distributer service, and router.

auth.go
```diff
@@ -411,10 +413,16 @@ func (x *Central) Login(identity, password string) (sessionkey string, token *To
...
 		x.Log.Infof("Login Authentication failed (%v) (%v)", identity, err)
+
+		// Send failed login to audit service
+		auditUserAction(identity, password, "Login Authentication failed")
 		return sessionkey, token, err
 	}
 	x.Log.Infof("Login authentication success (%v)", userId)
 
+	// Send success login to audit service
+	auditUserAction(identity, password, "Login Authentication success")
```
theunissens: We need to handle error in both cases. I am not sure if we should log it as an actual ERROR or WARNING if we cannot hit the audit service or their was any problem. Discuss how critical this is with @bmharper. I assume a Warning is enough, and if the user successfully logged in, things should flow as normal. If you only need to log a warning, you can just do it in the function itself.

bmharper: The error handling for auditUserAction should live inside auditUserAction, because the caller's behaviour doesn't change if the audit calls succeeds or fails.
I would remove those comments, since they're not adding any new information

auth.go
bmharper: I agree that it's too big. We should make a file called "audit.go", with this stuff inside it. Also, I suggested this morning to create an Auditor interface. It can be extremely simple - it just needs to take the parameters that auditUserAction takes. But we *must* do that, in order to preserve the property that authaus doesn't explicitly depend on any IMQS services.
It must also be optional - ie if the Auditor is null, then auditUserAction just returns without doing anything.

auth.go
```diff
@@ -444,6 +452,39 @@ func (x *Central) Login(identity, password string) (sessionkey string, token *To
...
 }
 
+func auditUserAction(identity, password, actionDescription string) error {
+	auditServiceUrl := "http://localhost:2016/auth-logaction"
+	client := &http.Client{}
+
+	currentTime := time.Now().Unix()
+	currentLocation := time.Now().Location().String()
+
+	actionStr := []byte(fmt.Sprintf(`{"who": "User %v with password %v","did_what": "%v", "to_what": "IMQS Web V8", "at_time": %v, "context": {"location": "%v"}}`, identity, password, actionDescription, currentTime, currentLocation))
```
bmharper: Rather use a struct and marshall it to JSON. Doing inline JSON like this is fine for very very small things, like a ping response, but it's going to lead to trouble when you accept arbitrary strings as input. For example, if any of the input strings had a double quote, you would have invalid JSON. Just use json.marshal, it's much neater.

auth.go
```diff
@@ -444,6 +452,39 @@ func (x *Central) Login(identity, password string) (sessionkey string, token *To
...
+
+	req, err := http.NewRequest("POST", auditServiceUrl, bytes.NewBuffer(actionStr))
+	if err != nil {
+		return err
+	}
+
+	// Date header is required for inter-service interactions
+	req.Header.Add("Date", time.Now().UTC().Format("Mon, 02 Jan 2006 15:04:05 GMT"))
```
bmharper: I think you can use time.RFC1123 here as the format.

auth.go
```diff
@@ -444,6 +452,39 @@ func (x *Central) Login(identity, password string) (sessionkey string, token *To
...
+	if err != nil {
+		return err
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+
+	defer resp.Body.Close()
```
bmharper: Just do `resp.Body.Close()`

auth.go
```diff
@@ -444,6 +452,39 @@ func (x *Central) Login(identity, password string) (sessionkey string, token *To
 	return sessionkey, token, nil
 }
 
+func auditUserAction(identity, password, actionDescription string) error {
```
bmharper: I don't think this function should return an error - primarily because the caller will never DO anything with that error. So rather have a wrapper function, which calls the Auditor's `AuditUserAction` function, and then logs the failure there.

audit.go
```diff
@@ -0,0 +1,60 @@
...
+	actionContext := &ActionContext{}
+	actionContext.Location = time.Now().Location().String()
+	action.Context = actionContext
+
+	client := &http.Client{}
+	actionStr, _ := json.Marshal(action)
+
+	req, err := http.NewRequest("POST", x.AuditServiceUrl, bytes.NewBuffer(actionStr))
+	if err != nil {
+		x.Log.Infof("Error creating audit request: (%v)", err)
```
bmharper: If you have an error here (or in subsequent places in this function), you cannot just continue executing as though there is no error. If NewRequest returns an error, then you can expect 'req' to be null, so you would just end up crashing on the next line. Also, this is not log level Info, this is Error.

audit.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (x *Central) AuditUserAction(identity, password, actionDescription string) {
+	action := &Action{}
+	action.Who = identity
+	action.DidWhat = actionDescription
+	action.ToWhat = "IMQS Web V8"
+	action.AtTime = time.Now().Unix()
+
+	actionContext := &ActionContext{}
+	actionContext.Location = time.Now().Location().String()
```
bmharper: I don't think that adding the server's time zone is useful information in the logs. Rather just omit information that we don't have.

audit.go
```diff
@@ -0,0 +1,60 @@
...
+
+	client := &http.Client{}
+	actionStr, _ := json.Marshal(action)
+
+	req, err := http.NewRequest("POST", x.AuditServiceUrl, bytes.NewBuffer(actionStr))
+	if err != nil {
+		x.Log.Infof("Error creating audit request: (%v)", err)
+	}
+
+	req.Header.Add("Date", time.Now().UTC().Format("Mon, 02 Jan 2006 15:04:05 GMT"))
```
bmharper: Unless I'm mistaken, use `time.RFC1123`

config.go
```diff
@@ -184,6 +184,7 @@ type Config struct {
 	RoleGroupDB ConfigRoleGroupDB
 	LDAP        ConfigLDAP
 	UserStore   ConfigUserStoreDB
+	AuditServiceUrl string
```
bmharper: Use an editor that runs gofmt when you save. One can see that this code has been edited with an editor that doesn't do that. Every reasonable Go editor I know of supports this.

audit.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"net/http"
+	"time"
+
+	"github.com/IMQS/serviceauth"
+)
+
+type Auditor interface {
```
bmharper: This interface isn't actually used anywhere.
Create a concrete type that implements the Auditor interface, and talks to the IMQS auditor interface. The concrete type's name should reference the fact that it's the IMQS audit service, so something like ImqsAuditor

bmharper:
One more thing - make the audit service optional. If the URL is not specified, then make audit calls do nothing.

frontends.go
```diff
@@ -84,7 +84,8 @@ func HttpHandlerLogin(config *ConfigHTTP, central *Central, w http.ResponseWrite
 		return
 	}
 
-	if sessionkey, token, err := central.Login(identity, password); err != nil {
+	clientIp := r.Header.Get("X-Forwarded-For")
```
bmharper: Try first for `X-Forwarded-For`. If that is not present, then fall back to `RemoteAddr`

auth.go
```diff
@@ -408,21 +411,25 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
 // Create a new session. Returns a session key, which can be used in future to retrieve the token.
 // The internal session expiry is controlled with the member NewSessionExpiresAfter.
 // The session key is typically sent to the client as a cookie.
-func (x *Central) Login(identity, password string) (sessionkey string, token *Token, err error) {
+func (x *Central) Login(identity, password, clientIp string) (sessionkey string, token *Token, err error) {
```
bmharper: clientIp never ends up in a log file.

auth.go
```diff
@@ -408,21 +411,25 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
...
+		if x.Auditor != nil {
+			x.auditUserAction(identity, "Login failed: Invalid password.")
+		}
 		return sessionkey, token, err
 	}
-	x.Log.Infof("Login authentication success (%v)", userId)
 
-	// Send success login to audit service
-	x.AuditUserAction(identity, password, "Logged in successfully")
+	// Log successful login
```
bmharper: No need for this comment. It's very clear what the code is doing.

auth.go
```diff
@@ -408,21 +411,25 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
...
+func (x *Central) Login(identity, password, clientIp string) (sessionkey string, token *Token, err error) {
 	userId, identity, authErr := x.authenticate(identity, password)
 	if authErr != nil {
 		err = authErr
 		x.Stats.IncrementInvalidPasswords(x.Log)
-		x.Log.Infof("Login Authentication failed (%v) (%v)", identity, err)
 
-		// Send failed login to audit service
-		x.AuditUserAction(identity, password, "Failed to log in")
+		// Log failed login
```
bmharper: No need for this comment. The code is very clear.

auth.go
```diff
@@ -408,21 +411,25 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
...
+		x.Log.Infof("Login Authentication failed (%v) (%v)", identity, err)
+		if x.Auditor != nil {
+			x.auditUserAction(identity, "Login failed: Invalid password.")
```
bmharper: nitpick: Don't finish log messages with a period.

config.go
bmharper: this is still not formatted with gofmt

PR #27, opened 2017-11-29
  TI-2402 - User list Filter

stewartm83:
Please review.

### DIFF
--- a/db.go
+++ b/db.go
@@ -24,6 +24,13 @@ var (
 	nextUserId    UserId
 )
 
+type GetIdentitiesFlag int
+
+const (
+	GetIdentitiesFlagNone    GetIdentitiesFlag = 0
+	GetIdentitiesFlagDeleted GetIdentitiesFlag = 1 << (iota - 1)
+)
+
 type UserId int64
 
 type AuthUserType int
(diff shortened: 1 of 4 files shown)


db.go
```diff
@@ -24,6 +24,13 @@ var (
 	nextUserId    UserId
 )
 
+type GetIdentitiesFlag int
+
+const (
+	GetIdentitiesFlagNone    GetIdentitiesFlag = 1 << iota
```
bmharper: The none flag must be zero. Simplest thing I can come up with is this:
```go
const (
    GetIdentitiesFlagNone GetIdentitiesFlag = 0
    GetIdentitiesFlagDeleted   = 1 << (iota - 1)
)
```

db.go
```diff
@@ -287,15 +295,16 @@ func (x *dummyUserStore) RenameIdentity(oldEmail, newEmail string) error {
...
-func (x *dummyUserStore) GetIdentities() ([]AuthUser, error) {
+func (x *dummyUserStore) GetIdentities(getIdentitiesFlag GetIdentitiesFlag) ([]AuthUser, error) {
 	x.usersLock.RLock()
 	defer x.usersLock.RUnlock()
 
 	list := []AuthUser{}
 	for _, v := range x.users {
-		if v.archived == false {
-			list = append(list, AuthUser{v.userId, v.email, v.username, v.firstname, v.lastname, v.mobilenumber, v.telephonenumber, v.remarks, v.created, v.createdby, v.modified, v.modifiedby, v.authUserType})
+		if getIdentitiesFlag != GetIdentitiesFlagDeleted && v.archived == true {
```
bmharper: This is a flag, so always use a binary AND to check for the presence of a flag.

db_sql.go
```diff
@@ -339,8 +339,14 @@ func (x *sqlUserStoreDB) RenameIdentity(oldIdent, newIdent string) error {
 	}
 }
 
-func (x *sqlUserStoreDB) GetIdentities() ([]AuthUser, error) {
-	rows, err := x.db.Query("SELECT userid, email, username, firstname, lastname, mobile, phone, remarks, created, createdby, modified, modifiedby, authusertype FROM authuserstore WHERE (archived = false OR archived IS NULL)")
+func (x *sqlUserStoreDB) GetIdentities(getIdentitiesFlag GetIdentitiesFlag) ([]AuthUser, error) {
+	var sqlStatement string
+	if getIdentitiesFlag == GetIdentitiesFlagDeleted {
```
bmharper: Use binary and to check for field presence.

Also, make the common part of the query a common string. And only add that final "WHERE" clause if the flag is not present.

RoanBrand: `if (getIdentitiesFlag && GetIdentitiesFlagDeleted) `

bmharper: Don't listen to him!
`if getIdentitiesFlag & GetIdentitiesFlagDeleted !=0`

bmharper:
Looks good

PR #29, opened 2018-02-20
TI-2400 - Audit Trail - Access logging

stewartm83:
Please review

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -204,8 +204,17 @@ func (x *CentralStats) IncrementLogout(logger *log.Logger) {
 	x.IncrementAndLog("logout", &x.Logout, logger)
 }
 
+type AuditActionType string
+
+const (
+	AuditActionUserAuthentication AuditActionType = "User authentication"
+	AuditActionUserCreated                        = "User created"
+	AuditActionUserUpdated                        = "User updated"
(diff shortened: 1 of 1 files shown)


auth.go
```diff
@@ -205,7 +205,7 @@ func (x *CentralStats) IncrementLogout(logger *log.Logger) {
 }
 
 type Auditor interface {
-	AuditUserAction(identity, clientIp, action string)
```
bmharper: Make action type an enum, even if it has only one type in it. Go doesn't have an enum type, but there's an idiomatic way to do enums.

auth.go
```diff
@@ -418,13 +418,13 @@ func (x *Central) Login(username, password, clientIp string) (sessionkey string,
 		x.Stats.IncrementInvalidPasswords(x.Log)
 
 		x.Log.Infof("Login Authentication failed (%v) (%v)", username, err)
-		x.auditUserAction(username, clientIp, "Login failed: Invalid password")
+		x.auditUserAction(username, clientIp, "Login failed: Invalid password", 1) // 1 - User authentication, eventually we will have many types and we will make this parameter generic
```
bmharper: Once the action type is an enum, we can get rid of the long comments.

PR #30, opened 2018-03-06
TI-2400 - Log user changes audit actions

stewartm83:
Please review.


bmharper:
Would it not be better if these audits happened in the HTTP API functions? The fact that we're missing WHO performed all of these actions, as well as the IP address of the actor, is disturbing.

bmharper:
Looks good. But I'd like to wait for the accompanying imqsauth changes before we merge this change.

PR #33, opened 2018-07-02
TI-2869 - Persist password change history

stewartm83:
Please review


db_sql.go
```diff
@@ -123,22 +126,71 @@ func (x *sqlUserStoreDB) ResetPasswordFinish(userId UserId, token string, passwo
...
+	return nil
+}
+
+func (x *sqlUserStoreDB) verifyPasswordArchived(userId UserId, password string) bool {
+	archivedPasswords, err := x.getArchivedPasswordsForUser(userId)
+	if err != nil {
+		return false
+	}
+	for _, hash := range archivedPasswords {
+		if verifyAuthausHash(password, hash) {
```
RoanBrand: A better way may be to calculate the hash first and then checking if it is in the archive table, instead of getting the last 15 passwords and checking each of them server side.

auth.go
```diff
@@ -44,6 +44,7 @@ var (
 	ErrInvalidSessionToken  = errors.New("Invalid session token")
 	ErrInvalidPasswordToken = errors.New("Invalid password token")
 	ErrPasswordTokenExpired = errors.New("Password token has expired")
+	ErrInvalidPastPassword  = errors.New("Invalid old password")
```
RoanBrand: Better: `Invalid: Previous used password`

db_sql.go
```diff
@@ -123,22 +126,71 @@ func (x *sqlUserStoreDB) ResetPasswordFinish(userId UserId, token string, passwo
...
 }
 
+func (x *sqlUserStoreDB) archivePassword(tx *sql.Tx, userId UserId) error {
+	if _, err := tx.Exec(`INSERT INTO authpwdarchive (userid, password) SELECT userid, password FROM authuserpwd WHERE userId = $1`, userId); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (x *sqlUserStoreDB) verifyPasswordArchived(userId UserId, password string) bool {
```
bmharper: The word 'verify' doesn't tell us what the function's return value means. It could rather be named something like `hasPasswordBeenUsedBefore`

db_sql.go
```diff
@@ -67,12 +67,15 @@ func (x *sqlUserStoreDB) Authenticate(identity, password string) error {
 
 func (x *sqlUserStoreDB) SetPassword(userId UserId, password string) error {
 	if tx, etx := x.db.Begin(); etx == nil {
-		if eupdate := x.setPasswordInternal(tx, userId, password); eupdate == nil {
-			return tx.Commit()
-		} else {
+		if used := x.verifyPasswordArchived(userId, password); used {
```
bmharper: No need for the `used` variable here. Just write `if x.hasPasswordBeenUsedBefore()`

db_sql.go
bmharper: @RoanBrand One actually needs to do it like this, because you can't compute the hash without the salt, and the salt is stored inside the blob that we refer to as the 'password hash'

db_sql.go
```diff
@@ -807,6 +859,16 @@ func createMigrations() []migration.Migrator {
...
 			ADD COLUMN modified TIMESTAMP,
 			ADD COLUMN modifiedby BIGINT;`,
+
+		// 10. Archive passwords
+		`
+		ALTER TABLE authuserstore  ALTER COLUMN modified SET DEFAULT NOW();	
+		ALTER TABLE authuserpwd  ADD COLUMN created TIMESTAMP DEFAULT NOW();
+		ALTER TABLE authuserpwd  ADD COLUMN updated TIMESTAMP DEFAULT NOW();	
+
+		CREATE TABLE authpwdarchive (id BIGSERIAL PRIMARY KEY, userid BIGINT NOT NULL, password VARCHAR NULL, created TIMESTAMP DEFAULT NOW());
```
bmharper: We don't usually include the explicit "NULL" that is present here in "password VARCHAR NULL".

db_sql.go
```diff
@@ -123,22 +126,71 @@ func (x *sqlUserStoreDB) ResetPasswordFinish(userId UserId, token string, passwo
...
+	passwords := make([]string, 0)
+	rows, err := x.db.Query("SELECT password FROM authpwdarchive WHERE userid = $1 ORDER BY created DESC LIMIT 15", userId)
+	if err != nil {
+		return passwords, err
+	}
+	defer rows.Close()
+
+	for rows.Next() {
+		var passwordDbHash string
+		errScan := rows.Scan(&passwordDbHash)
```
bmharper: I see from the migration that passwordDbHash is allowed to be NULL, but this code isn't built to deal with nulls. You'll need to do use sql.NullString instead for that.
BUT.. if you actually meant to make the archived password field "NOT NULL", then obviously the migration needs to be changed.

PR #34, opened 2018-07-19
TI-2878 - Add ability to expire user passwords

stewartm83:


db.go
```diff
@@ -57,10 +65,11 @@ func (u AuthUserType) CanRenameIdentity() bool {
...
-	SetPassword(userId UserId, password string, checkPasswordReuse bool) error                       // This sets the password to a user account
+	Authenticate(identity, password string, authTypeCheck AuthCheck) (string, error) // Return nil error if the username and password are correct, otherwise one of ErrIdentityAuthNotFound or ErrInvalidPassword
+	SetPassword(userId UserId, password string, authTypeCheck AuthCheck) error       // This sets the password to a user account
+	SetConfig(passwordExpiry time.Duration) error
 	ResetPasswordStart(userId UserId, expires time.Time) (string, error)                             // Create a one-time token that can be used to reset the password with a subsequent call to ResetPasswordFinish
-	ResetPasswordFinish(userId UserId, token string, password string, checkPasswordReuse bool) error // Check that token matches the last one generated by ResetPasswordStart, and if so, call SetPassword
+	ResetPasswordFinish(userId UserId, token string, password string, authTypeCheck AuthCheck) error // Check that token matches the last one generated by ResetPasswordStart, and if so, call ResetPasswordFinish
```
RoanBrand: but this is `ResetPasswordFinish`? Is this comment correct?

db.go
```diff
@@ -184,19 +197,24 @@ func (x *dummyUserStore) Authenticate(identity, password string) (er error) {
...
+	x.passswordExpiry = passwordExpiry
+	return nil
+}
+
+func (x *dummyUserStore) SetPassword(userId UserId, password string, authTypeCheck AuthCheck) error {
 	x.usersLock.Lock()
 	defer x.usersLock.Unlock()
 	if user, exists := x.users[userId]; exists && user.authUserType.CanSetPassword() {
-		if checkPasswordReuse && user.password == password {
+		if authTypeCheck&AuthCheckNonPasswordReuse == 1 && user.password == password {
```
RoanBrand: Firstly it does not look like this file is formatted.

And you went through the trouble of creating a new type but not using it like so:
`authTypeCheck == AuthCheckNonPasswordReuse`

auth.go
```diff
@@ -403,24 +411,27 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
...
-			t.UserId = userId
+			t.Identity = user.getIdentity()
+			t.UserId = user.UserId
 			t.Permit = *permit
 			x.Stats.IncrementGoodOnceOffAuth(x.Log)
-			x.Log.Infof("Once-off auth successful (%v)", userId)
+			x.Log.Infof("Once-off auth successful (%v)", user.UserId)
+			if authMsg != "" {
+				t.Message = authMsg
+			}
```
RoanBrand: Just `t.Message = authMsg` because it will do exactly the same. 

And you can shove it in at `t`s creation time like i mentioned above.

db_sql.go
```diff
@@ -43,31 +43,44 @@ const (
...
-		return nil
-	} else {
-		return ErrInvalidPassword
+	if !verifyAuthausHash(password, dbHash) {
+		return "", ErrInvalidPassword
+	}
+
+	if x.passwordExpiry != 0 && authTypeCheck&AuthCheckPasswordExpired == 2 {
+		if lastUpdated.Add(x.passwordExpiry).Before(time.Now()) {
+			return ErrPasswordExpired.Error(), nil
```
RoanBrand: This is not done right IMO. You can return this error as the second argument (which is the standard Go way) and then react differently to it upstream.

db.go
```diff
@@ -31,6 +31,14 @@ const (
 	GetIdentitiesFlagDeleted GetIdentitiesFlag = 1 << (iota - 1)
 )
 
+type AuthCheck int
+
+const (
+	AuthCheckNone             AuthCheck = 0
+	AuthCheckNonPasswordReuse AuthCheck = 1 << (iota - 1)
+	AuthCheckPasswordExpired
+)
```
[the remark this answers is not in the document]
stewartm83: The idea was to use bit-wise comparison instead of standard int's. You can see this by adding an extra item to the enum.

```
const (
	AuthCheckNone             AuthCheck = 0
	AuthCheckNonPasswordReuse AuthCheck = 1<<(iota -1)
	AuthCheckPasswordExpired
        AuthCheckPasswordNext
)
```
AuthCheckPasswordNext will be different in these scenarios.

db.go
bmharper: Right way to check for flags is: `!!(authTypeCheck&AuthCheckNonPasswordReuse)`. That works in C++. Not sure about Go. You might have to do `(authTypeCheck&AuthCheckNonPasswordReuse) != 0`

db.go
```diff
@@ -31,6 +31,14 @@ const (
 	GetIdentitiesFlagDeleted GetIdentitiesFlag = 1 << (iota - 1)
 )
 
+type AuthCheck int
+
+const (
+	AuthCheckNone             AuthCheck = 0
+	AuthCheckNonPasswordReuse AuthCheck = 1 << (iota - 1)
```
bmharper: If `AuthCheckNonPasswordReuse` could be turned into a positive flag, aka `AuthCheckPasswordReuse`, that would be cleaner

db.go
```diff
@@ -217,22 +232,21 @@ func (x *dummyUserStore) ResetPasswordStart(userId UserId, expires time.Time) (s
...
-func (x *dummyUserStore) ResetPasswordFinish(userId UserId, token string, password string, checkPasswordReuse bool) error {
+func (x *dummyUserStore) ResetPasswordFinish(userId UserId, token string, password string, authTypeCheck AuthCheck) error {
 	x.usersLock.Lock()
 	defer x.usersLock.Unlock()
 	if user, exists := x.users[userId]; exists && !user.archived && user.authUserType.CanSetPassword() {
 		if err := verifyPasswordResetToken(token, user.passwordResetToken); err != nil {
 			return err
 		}
-		if checkPasswordReuse && user.password == password {
+		if authTypeCheck&AuthCheckNonPasswordReuse == 1 && user.password == password {
```
bmharper: same flag check issue

db_sql.go
```diff
@@ -43,31 +43,44 @@ const (
...
 
-	if verifyAuthausHash(password, dbHash) {
-		return nil
-	} else {
-		return ErrInvalidPassword
+	if !verifyAuthausHash(password, dbHash) {
+		return "", ErrInvalidPassword
+	}
+
+	if x.passwordExpiry != 0 && authTypeCheck&AuthCheckPasswordExpired == 2 {
```
bmharper: flag check issue

db_sql.go
```diff
@@ -43,31 +43,44 @@ const (
...
+		}
 	}
+	return "", nil
 }
 
-func (x *sqlUserStoreDB) SetPassword(userId UserId, password string, checkPasswordReuse bool) error {
+func (x *sqlUserStoreDB) SetPassword(userId UserId, password string, authTypeCheck AuthCheck) error {
 	if tx, etx := x.db.Begin(); etx == nil {
-		if checkPasswordReuse && x.hasPasswordBeenUsedBefore(userId, password) {
+		if authTypeCheck&AuthCheckNonPasswordReuse == 1 && x.hasPasswordBeenUsedBefore(userId, password) {
```
bmharper: flag check issue

db_sql.go
```diff
@@ -138,7 +151,7 @@ func (x *sqlUserStoreDB) ResetPasswordFinish(userId UserId, token string, passwo
 			return everify
 		}
 
-		if checkPasswordReuse && x.hasPasswordBeenUsedBefore(userId, password) {
+		if authTypeCheck&AuthCheckNonPasswordReuse == 1 && x.hasPasswordBeenUsedBefore(userId, password) {
```
bmharper: flag check issue

auth.go
```diff
@@ -438,43 +452,51 @@ func (x *Central) Login(username, password string) (sessionkey string, token *To
...
-	token.UserId = userId
+	token = &Token{
+		Permit:   *permit,
+		Identity: user.getIdentity(),
+		UserId:   user.UserId,
+	}
+
+	userPasswordExpiry := user.PasswordModifiedDate.Add(x.PasswordExpiresAfter)
+	sessionExpiry := time.Now().Add(x.NewSessionExpiresAfter)
+	if userPasswordExpiry.After(time.Now()) && userPasswordExpiry.Before(sessionExpiry) {
```
bmharper: I don't understand the role of `userPasswordExpiry.After(time.Now())`. As far as I can tell, it shouldn't be there.

db.go
```diff
@@ -31,6 +31,14 @@ const (
 	GetIdentitiesFlagDeleted GetIdentitiesFlag = 1 << (iota - 1)
 )
 
+type AuthCheck int
```
bmharper: Sorry to create more issues, but these things MUST be in separate enums. They don't share any functionality.
One of them is AuthCheck, and the other one is PasswordEnforcement.
Also, I would rename AuthCheckNone to AuthCheckDefault, because None implies that we're not even checking the password.

db_sql.go
```diff
@@ -166,20 +179,27 @@ func (x *sqlUserStoreDB) archivePassword(tx *sql.Tx, userId UserId) error {
 }
 
 func (x *sqlUserStoreDB) hasPasswordBeenUsedBefore(userId UserId, password string) bool {
-	archivedPasswords, err := x.getArchivedPasswordsForUser(userId)
+	pastPasswords, err := x.getPastPasswordsForUser(userId)
```
bmharper: This would read a lot easier if it was `recentPasswords` and `getRecentPasswordsForUser`, instead of `pastpasswords`

bmharper:
LGTM

PR #35, opened 2018-08-03
Fix: Ldap users session expires immediately.

stewartm83:
Please review.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -410,7 +410,7 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
 		x.Stats.IncrementEmptyIdentities(x.Log)
 		return nil, ErrIdentityEmpty
 	}
-	user, eAuth := x.authenticate(identity, password, AuthCheckDefault)
+	user, eAuth := x.authenticate(identity, password)
 	if eAuth == nil {
 		if permit, ePermit := x.permitDB.GetPermit(user.UserId); ePermit == nil {
 			t := &Token{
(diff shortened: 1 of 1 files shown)


auth.go
```diff
@@ -472,14 +472,11 @@ func (x *Central) Login(username, password string) (sessionkey string, token *To
...
+
 	sessionExpiry := time.Now().Add(x.NewSessionExpiresAfter)
-	if x.PasswordExpiresAfter != 0 {
-		userPasswordExpiry := user.PasswordModifiedDate.Add(x.PasswordExpiresAfter)
-		if userPasswordExpiry.Before(sessionExpiry) {
-			token.Expires = userPasswordExpiry
-		} else {
-			token.Expires = sessionExpiry
-		}
+	userPasswordExpiry := user.PasswordModifiedDate.Add(x.PasswordExpiresAfter)
```
bmharper: Why did you remove the check for `x.PasswordExpiresAfter != 0`. That seems like a valid thing to keep.

auth.go
```diff
@@ -472,10 +472,11 @@ func (x *Central) Login(username, password string) (sessionkey string, token *To
 		Identity: user.getIdentity(),
 		UserId:   user.UserId,
 	}
+
 	sessionExpiry := time.Now().Add(x.NewSessionExpiresAfter)
-	if x.PasswordExpiresAfter != 0 {
+	if x.PasswordExpiresAfter != 0 && x.ldap == nil {
```
stewartm83: @bmharper can we, for the moment, just check whether it is an ldap call and skip password expiry check?

bmharper: I definitely wouldn't say "for the moment". Any code that gets committed is likely to remain there forever. But will this approach work for a hybrid of ldap and IMQS users in the system? This is actually quite a common scenario.

PR #36, opened 2018-08-21
TI-3017 - Rename audit actions to make them generic

stewartm83:
Please review


RoanBrand:
was this really necessary?

stewartm83:
Initially I thought we were going to log changes to user profiles only, but since now we have to log changes done to groups as well, I think it makes sense that the names are more generic.

PR #38, opened 2018-12-03
TI-3140 - Log changes to Ldap users to user audit trail and add missing details

stewartm83:
Please review

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -5,6 +5,7 @@ import (
 	"crypto/rand"
 	"crypto/subtle"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strconv"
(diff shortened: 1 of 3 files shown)


auth.go
```diff
@@ -566,6 +565,21 @@ func (x *Central) MergeTick() {
 	}
 }
 
+func serializeUserInfo(userId UserId, username string) []byte {
```
bmharper: This function needs a better name, such as `userInfoToAuditTrailJSON()`
Also, why return `[]byte`, if you always cast it to a `string`? Rather just return `string`.
Also, why have two parameters to the function, instead of just taking an `*AuthUser` object?

bmharper:
LGTM. I must say it's a bit dirty that we're embedding the audit train JSON in here, but it's a tolerable hack.

PR #42, opened 2019-05-28
TI-3497: Fix for ldap user update loop

FritzOnFire:
A ldap user with a space at the end of his email caused the system to update him
during every ldap merge as the space at the end of the email would mysteriously
dissapear after the user structs address was taken. We trim the spaces of the
emails now by default. Add exstra loging for ldap merge user update.

### DIFF
--- a/db.go
+++ b/db.go
@@ -118,21 +118,21 @@ type SessionDB interface {
 }
 
 type AuthUser struct {
-	UserId               UserId
-	Email                string
-	Username             string
-	Firstname            string
-	Lastname             string
-	Mobilenumber         string
-	Telephonenumber      string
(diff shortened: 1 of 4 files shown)


db.go
```diff
@@ -118,21 +118,21 @@ type SessionDB interface {
...
-	Telephonenumber      string
-	Remarks              string
-	Created              time.Time
-	CreatedBy            UserId
-	Modified             time.Time
-	ModifiedBy           UserId
-	Type                 AuthUserType
-	Archived             bool
-	PasswordModifiedDate time.Time
+	UserId               UserId       `json:"userid"`
```
bmharper: Why not camelCase?

PR #43, opened 2019-06-13
Fix for row scan failing on a null password

FritzOnFire:
This would return an ErrIdentityAuthNotFound which was missleading as the user
does indeed exsit at this point.

### DIFF
--- a/db_sql.go
+++ b/db_sql.go
@@ -59,13 +59,18 @@ func (x *sqlUserStoreDB) Authenticate(identity, password string, authTypeCheck A
 	}
 
 	row = x.db.QueryRow(`SELECT password, updated FROM authuserpwd WHERE userid = $1`, userId)
-	dbHash := ""
+	var dbHash sql.NullString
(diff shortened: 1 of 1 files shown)


db_sql.go
```diff
@@ -59,13 +59,18 @@ func (x *sqlUserStoreDB) Authenticate(identity, password string, authTypeCheck A
...
 	row = x.db.QueryRow(`SELECT password, updated FROM authuserpwd WHERE userid = $1`, userId)
-	dbHash := ""
+	var dbHash sql.NullString
 	var lastUpdated time.Time
 	if err := row.Scan(&dbHash, &lastUpdated); err != nil {
 		return ErrIdentityAuthNotFound
 	}
 
-	if !verifyAuthausHash(password, dbHash) {
+	pHash := ""
```
bmharper: Remember to add a comment here explaining how the password field ends up being null.

bmharper:
LGTM

PR #44, opened 2019-06-28
TI-3416 - Lock account after 3 failed login attempts

nkosimarz:
- Lock account after 3 login attempts
- Save account locked status in DB

### DIFF
--- a/all_test.go
+++ b/all_test.go
@@ -19,7 +19,6 @@ NOTE: Some of these tests stress concurrency, so you must run them with at least
 TODO: Add test that verifies that SetPassword does not create an identity if that identity does not already exist
 
 Create a test Postgres database:
-	create role authaus_test login password 'authaus_test';
 	create database authaus_test owner = authaus_test;
 
 Suggested test runs that you should do:
@@ -264,10 +263,10 @@ func setup(t *testing.T) *Central {
 func setupLdap(t *testing.T) *Central {
 	central := getCentral(t)
 
-	if isBackendLdapTest() {	
+	if isBackendLdapTest() {
 		central.ldap = NewAuthenticator_LDAP(&conx_ldap)
 		central.MergeTick()
-		
+
 		// Setup permissions
 		user, err := central.userStore.GetUserFromIdentity(imqsLdapIdentity)
 		if err != nil {
@@ -325,6 +324,18 @@ func isPrefix(prefix, str string) bool {
 	return strings.Index(str, prefix) == 0
 }
 
+type TestLockableUser struct {
+	enableLocking bool
+}
+
+func (x TestLockableUser) IsLockable(identity string) (bool, error) {
+	return x.enableLocking, nil
+}
+
+func (x Central) AuditUserAction(identity, item, context string, auditActionType AuditActionType) {
+	return
+}
+
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 // Integrated LDAP Tests
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
@@ -342,13 +353,13 @@ func TestIntegratedLdapAuthenticateFailure(t *testing.T) {
 	joePermit := setupPermit()
 	c.permitDB.SetPermit(user.UserId, &joePermit)
 
-	if _, _, err := c.Login(imqsLdapIdentity, "invalidpassword"); err != nil {
+	if _, _, err := c.Login(imqsLdapIdentity, "invalidpassword", ""); err != nil {
 		if !strings.Contains(err.Error(), ErrInvalidPassword.Error()) {
 			t.Fatalf("Login should have failed with invalid password, but error was: %v", err)
 		}
 	}
 
-	if _, _, err := c.Login("somerandomusername", "invalidpassword"); err != nil {
+	if _, _, err := c.Login("somerandomusername", "invalidpassword", ""); err != nil {
 		if !strings.Contains(err.Error(), ErrIdentityAuthNotFound.Error()) {
 			t.Fatalf("Login should have failed with identity not found, but error was: %v", err)
 		}
@@ -370,7 +381,7 @@ func TestIntegratedLdapLoginWithAnonymousBind(t *testing.T) {
 	c.permitDB.SetPermit(user.UserId, &joePermit)
 
 	// Note, we are logging in with no password
-	if _, _, err := c.Login(imqsLdapIdentity, ""); err != nil {
+	if _, _, err := c.Login(imqsLdapIdentity, "", ""); err != nil {
 		if !strings.Contains(err.Error(), ErrInvalidPassword.Error()) {
 			t.Fatalf("Login should have failed with invalid password, but error was: %v", err)
 		}
@@ -394,7 +405,7 @@ func TestIntegratedLdapConnectionRecovery(t *testing.T) {
 	joePermit := setupPermit()
 	c.permitDB.SetPermit(user.UserId, &joePermit)
 
-	if _, _, err := c.Login(imqsLdapIdentity, imqsLdapPwd); err != nil {
+	if _, _, err := c.Login(imqsLdapIdentity, imqsLdapPwd, ""); err != nil {
 		t.Fatalf("Login should have succeeded, but error was : %v", err)
 	}
 
@@ -406,7 +417,7 @@ func TestIntegratedLdapConnectionRecovery(t *testing.T) {
 	conx_ldap.LdapHost = host
 	c.MergeTick()
 
-	if _, _, err := c.Login(imqsLdapIdentity, imqsLdapPwd); err != nil {
+	if _, _, err := c.Login(imqsLdapIdentity, imqsLdapPwd, ""); err != nil {
 		t.Fatalf("Login should have succeeded, but error was : %v", err)
 	}
 }
@@ -427,7 +438,7 @@ func TestIntegratedLdapMergeLoad(t *testing.T) {
 		for {
 			select {
 			case <-ticker.C:
-				_, _, err = c.Login(imqsLdapIdentity, imqsLdapPwd)
+				_, _, err = c.Login(imqsLdapIdentity, imqsLdapPwd, "")
 				if err != nil {
 					t.Errorf("Login failed %v", err)
 				}
@@ -467,14 +478,14 @@ func TestAuthLdapMerge(t *testing.T) {
 	tomhPermit := setupPermit()
 	c.permitDB.SetPermit(testLdapUser.UserId, &tomhPermit)
 
-	if _, _, err := c.Login(testLdapIdentity, testLdapPwd); err != nil {
+	if _, _, err := c.Login(testLdapIdentity, testLdapPwd, ""); err != nil {
 		t.Fatalf("Login should have succeeded, but error was : %v", err)
 	}
 
 	ldapTest.RemoveLdapUser(testLdapIdentity)
 	c.MergeTick()
 
-	if _, _, err := c.Login(testLdapIdentity, testLdapPwd); err == nil {
+	if _, _, err := c.Login(testLdapIdentity, testLdapPwd, ""); err == nil {
 		t.Fatalf("Login should not have succeeded")
 	}
 
@@ -633,7 +644,7 @@ func TestAuthLdapIMQSUserToLDAPUserConversion(t *testing.T) {
 	// Test IMQS user login.
 	permit := setupPermit()
 	c.permitDB.SetPermit(peterUserId, &permit)
-	_, _, err = c.Login(newEmail, newIMQSPwd)
+	_, _, err = c.Login(newEmail, newIMQSPwd, "")
 	if err != nil {
 		t.Fatalf("Login should have succeeded, but error was : %v", err)
 	}
@@ -642,18 +653,18 @@ func TestAuthLdapIMQSUserToLDAPUserConversion(t *testing.T) {
 	// should no longer work. However, login with the LDAP user should work.
 	c.MergeTick()
 
-	_, _, err = c.Login(newEmail, newIMQSPwd)
+	_, _, err = c.Login(newEmail, newIMQSPwd, "")
 	if err == nil {
 		t.Fatalf("Login should have failed")
 	}
 
 	// Login with both email address and username of the LDAP account to make sure the account is
 	// one and the same.
-	_, emailLoginToken, err := c.Login(newEmail, newLDAPPwd)
+	_, emailLoginToken, err := c.Login(newEmail, newLDAPPwd, "")
 	if err != nil {
 		t.Fatalf("Login should have succeeded, but error was : %v", err)
 	}
-	_, usernameLoginToken, err := c.Login(newLDAPUsername, newLDAPPwd)
+	_, usernameLoginToken, err := c.Login(newLDAPUsername, newLDAPPwd, "")
 	if err != nil {
 		t.Fatalf("Login should have succeeded, but error was : %v", err)
 	}
@@ -702,14 +713,14 @@ func TestAuthLdapUsernamesAndEmailsWithEmptyStringsShouldNotMerge(t *testing.T) 
 
 	permit := setupPermit()
 	c.permitDB.SetPermit(imqsUserId, &permit)
-	_, _, err = c.Login(imqsUsername, imqsPassword)
+	_, _, err = c.Login(imqsUsername, imqsPassword, "")
 	if err != nil {
 		t.Fatalf("Login should have succeeded, but error was : %v", err)
 	}
 
 	// This merge should have zero affect, no merge on this user should take place
 	c.MergeTick()
-	_, _, err = c.Login(imqsUsername, imqsPassword)
+	_, _, err = c.Login(imqsUsername, imqsPassword, "")
 	if err != nil {
 		t.Fatalf("Login should have succeeded. This probably means the merge took place, and should not have. Error was: %v", err)
 	}
@@ -735,14 +746,14 @@ func TestAuthLdapUsernamesAndEmailsWithEmptyStringsShouldNotMerge(t *testing.T) 
 	ldapTest.AddLdapUser(emptyStringUsername, "", ldapEmail, "", "", "")
 
 	c.permitDB.SetPermit(imqsUserId, &permit)
-	_, _, err = c.Login(imqsEmail, imqsPassword)
+	_, _, err = c.Login(imqsEmail, imqsPassword, "")
 	if err != nil {
 		t.Fatalf("Login should have succeeded, but error was : %v", err)
 	}
 
 	// This merge should have zero affect, no merge on this user should take place
 	c.MergeTick()
-	_, _, err = c.Login(imqsEmail, imqsPassword)
+	_, _, err = c.Login(imqsEmail, imqsPassword, "")
 	if err != nil {
 		t.Fatalf("Login should have succeeded. This probably means the merge took place, and should not have. Error was: %v", err)
 	}
@@ -784,7 +795,7 @@ func TestAuthLoginCaseSensitivity(t *testing.T) {
 
 	const joeCapitalizedIdentity = "JOE@email.test"
 
-	_, token, e := c.Login(joeCapitalizedIdentity, joePwd)
+	_, token, e := c.Login(joeCapitalizedIdentity, joePwd, "")
 	if e != nil {
 		t.Fatalf("An unexpected error occurred: %v", e)
 	}
@@ -809,7 +820,7 @@ func TestAuthRenameIdentity(t *testing.T) {
 	}
 
 	// Succeed renaming 'joe' to 'sarah'
-	session, _, _ := c.Login(joeEmail, joePwd)
+	session, _, _ := c.Login(joeEmail, joePwd, "")
 	if _, err := c.GetTokenFromSession(session); err != nil {
 		t.Fatalf("Expected good login")
 	}
@@ -822,7 +833,7 @@ func TestAuthRenameIdentity(t *testing.T) {
 		t.Fatalf("All sessions for 'joe' should have been invalidated by rename, %s", err)
 	}
 
-	if _, _, err := c.Login("sarah", joePwd); err != nil {
+	if _, _, err := c.Login("sarah", joePwd, ""); err != nil {
 		t.Fatalf("Login as 'sarah' failed (%v)", err)
 	}
 }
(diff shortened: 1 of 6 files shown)


auth.go
```diff
@@ -436,12 +461,23 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
...
 	if authErr != nil {
+
 		err = authErr
-		x.Stats.IncrementInvalidPasswords(x.Log)
 
+		if err == ErrInvalidPassword {
+			x.Stats.IncrementInvalidPasswords(x.Log)
+			x.Stats.IncrementInvalidPasswordHistory(x.Log, username, address)
+			if isAdmin, groupErr := x.CheckAdmin(user.UserId); groupErr == nil {
+				if int(x.Stats.UserLoginAttempts[username]) >= 3 && x.EnableAccountLocking && !isAdmin {
```
bmharper: 3 is hardcoded. should use the configured limit.
Move the `x.EnableAccountLocking` check higher up, so that we don't even need to take the mutex lock if this functionality is not enabled.

auth.go
```diff
@@ -503,6 +540,24 @@ func (x *Central) authenticate(identity, password string) (AuthUser, error) {
 		authTypeCheck = AuthCheckDefault
 	}
 
+	InvalidPasswords := x.Stats.UserLoginAttempts[identity]
```
bmharper: `InvalidPasswords` -> `invalidPasswords`

auth.go
```diff
@@ -520,6 +575,30 @@ func (x *Central) authenticate(identity, password string) (AuthUser, error) {
 	}
 }
 
+func (x *Central) CheckAdmin(userId UserId) (bool, error) {
```
bmharper: `Check` is a classic no-no for function names. Rather call it `IsAdmin`.

I would split this into two functions. One of them should just be generic, and search for the user's presence in any group (ie pass the magic string "admin" in as a function parameter).

Also, put a clear comment here, saying that the word "admin" is the magic group name used by the imqsauth. This is hack, bringing in knowledge from the imqsauth package, but we feel it's OK for this particular edge case.

db_sql.go
```diff
@@ -58,6 +62,12 @@ func (x *sqlUserStoreDB) Authenticate(identity, password string, authTypeCheck A
 		return ErrIdentityAuthNotFound
 	}
 
+	var accountLocked bool
+	row = x.db.QueryRow(`SELECT accountlocked FROM authuserpwd WHERE userid = $1`, userId)
+	if err := row.Scan(&accountLocked); err != nil || accountLocked {
```
bmharper: What are you doing about NULLs here? Rather use sql.NullBool and make the behaviour explicit and crystal clear.
Also, just write `x.db.QueryRow().Scan()`

db_sql.go
```diff
@@ -296,6 +306,20 @@ func identityFromRow(row *sql.Row) error {
 	return ErrIdentityExists
 }
 
+func (x *sqlUserStoreDB) LockAccount(userId UserId) error {
+	if _, err := x.db.Exec(`UPDATE authuserpwd SET accountlocked = true WHERE userId = $1;`, userId); err != nil {
+		return err
+	}
+	return ErrAccountLocked
```
bmharper: This should not be an error.

db_sql.go
```diff
@@ -893,6 +919,8 @@ func createMigrations() []migration.Migrator {
 
 		CREATE TABLE authpwdarchive (id BIGSERIAL PRIMARY KEY, userid BIGINT NOT NULL, password VARCHAR NOT NULL, created TIMESTAMP DEFAULT NOW());
 		`,
+		// 11. Account lock
+		`ALTER TABLE authuserpwd ADD COLUMN accountlocked BOOLEAN;`,
```
bmharper: What's the policy with NULL? Does NULL mean locked or not locked? Do new entries in the table get it set to false or NULL?

db_sql.go
```diff
@@ -42,12 +42,16 @@ const (
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 type sqlUserStoreDB struct {
-	db             *sql.DB
-	passwordExpiry time.Duration
+	db                   *sql.DB
+	passwordExpiry       time.Duration
+	enableAccountLocking bool
```
bmharper: Does enableAccountLocking and maxFailedPasswords need to be here? I think we don't need this state inside the UserStore. I think it can just live at the higher levels (eg inside authaus.Central)

db.go
```diff
@@ -73,7 +73,7 @@ func (u AuthUserType) CanRenameIdentity() bool {
 type UserStore interface {
 	Authenticate(identity, password string, authTypeCheck AuthCheck) error                  // Return nil error if the username and password are correct, otherwise one of ErrIdentityAuthNotFound or ErrInvalidPassword
 	SetPassword(userId UserId, password string, enforceTypeCheck PasswordEnforcement) error // This sets the password to a user account
-	SetConfig(passwordExpiry time.Duration) error
+	SetConfig(passwordExpiry time.Duration, enableAccountLocking bool, maxFailedPasswords int) error
```
bmharper: See comments below -- I don't think we need this change to the the SetConfig function

config.go
```diff
@@ -174,6 +174,8 @@ type ConfigUserStoreDB struct {
 	DB                    DBConnection
 	DisablePasswordReuse  bool
 	PasswordExpirySeconds int
+	EnableAccountLocking  bool
```
bmharper: Don't think this is necessary

auth.go
```diff
@@ -794,6 +873,18 @@ func (x *Central) UpdateIdentity(user *AuthUser) error {
 	return nil
 }
 
+// Update a user in the AuthUserStore.
+func (x *Central) UnlockAccount(user *AuthUser) error {
```
bmharper: Since the function only uses the userID, rather make the parameter int64, instead of *AuthUser

db.go
```diff
@@ -82,6 +82,8 @@ type UserStore interface {
 	// TODO RenameIdentity was deprecated in May 2016, replaced by UpdateIdentity. We need to remove this once PCS team has made the necessary updates
 	RenameIdentity(oldIdent, newIdent string) error                        // Rename an identity. Returns ErrIdentityAuthNotFound if oldIdent does not exist. Returns ErrIdentityExists if newIdent already exists.
 	GetUserFromIdentity(identity string) (AuthUser, error)                 // Gets the user object from the identity supplied
+	LockAccount(userId UserId) error                                       // Locks account after 3 failed log in attempts
+	UnlockAccount(userId UserId) error                                     // Locks account after 3 failed log in attempts
```
FritzOnFire: I don't think this is the right comment...

auth.go
```diff
@@ -436,12 +461,23 @@ func (x *Central) GetTokenFromIdentityPassword(identity, password string) (*Toke
...
 // The session key is typically sent to the client as a cookie.
-func (x *Central) Login(username, password string) (sessionkey string, token *Token, err error) {
+func (x *Central) Login(username, password string, address string) (sessionkey string, token *Token, err error) {
 	user, authErr := x.authenticate(username, password)
 	if authErr != nil {
+
 		err = authErr
-		x.Stats.IncrementInvalidPasswords(x.Log)
 
+		if err == ErrInvalidPassword {
```
bmharper: This check cannot be performed here, because the user doesn't need to try to Login in order to check if the password is correct. He can also try to performance a once-off authorization. So all of this account locking stuff needs to happen inside authorize().

db.go
```diff
@@ -330,20 +338,28 @@ func (x *dummyUserStore) GetIdentities(getIdentitiesFlag GetIdentitiesFlag) ([]A
...
 			continue
 		}
-		list = append(list, AuthUser{v.userId, v.email, v.username, v.firstname, v.lastname, v.mobilenumber, v.telephonenumber, v.remarks, v.created, v.createdby, v.modified, v.modifiedby, v.authUserType, v.archived, v.passwordModifiedDate})
+		list = append(list, AuthUser{v.userId, v.email, v.username, v.firstname, v.lastname, v.mobilenumber, v.telephonenumber, v.remarks, v.created, v.createdby, v.modified, v.modifiedby, v.authUserType, v.archived, v.passwordModifiedDate, v.accountLocked})
 	}
 	return list, nil
 }
 
+func (x *dummyUserStore) LockAccount(userId UserId) error {
+	return nil
```
FritzOnFire: I don't know what the value would be... but this should probably be:
```go
user := x.users[userId]
if user.archived {
    return // some Error
}

user->accountLocked = false
return nil
```

db.go
```diff
@@ -330,20 +338,28 @@ func (x *dummyUserStore) GetIdentities(getIdentitiesFlag GetIdentitiesFlag) ([]A
...
+}
+
+func (x *dummyUserStore) UnlockAccount(userId UserId) error {
+	return nil
```
FritzOnFire: I don't know what the value would be... but this should probably be:
```go
user := x.users[userId]
if user.archived {
    return // some Error
}

user->accountLocked = true
return nil
```

db.go
```diff
@@ -82,6 +82,8 @@ type UserStore interface {
 	// TODO RenameIdentity was deprecated in May 2016, replaced by UpdateIdentity. We need to remove this once PCS team has made the necessary updates
 	RenameIdentity(oldIdent, newIdent string) error                        // Rename an identity. Returns ErrIdentityAuthNotFound if oldIdent does not exist. Returns ErrIdentityExists if newIdent already exists.
 	GetUserFromIdentity(identity string) (AuthUser, error)                 // Gets the user object from the identity supplied
+	LockAccount(userId UserId) error                                       // Locks account after 3 failed log in attempts
```
bmharper: Comment is inaccurate - 3 is not hardcoded.
Also, both comments are the same.

bmharper:
One more thing - I don't think the Account Locked action ever makes it into the audit log

auth.go
```diff
@@ -179,6 +182,32 @@ func (x *CentralStats) IncrementAndLog(name string, val *uint64, logger *log.Log
...
+	if count < 5 || isPowerOf2(count) {
+		logger.Infof("%v has %v invalid password attempts from %v", username, count, clientIPAddress)
+	}
+}
+
+func (x *CentralStats) ResetInvalidPasswordHistory(logger *log.Logger, username string, clientIPAddress string) {
+	x.loginLock.Lock()
+	defer x.loginLock.Unlock()
+	oldCount := x.UserLoginAttempts[username]
+	if x.UserLoginAttempts == nil {
```
bmharper: The nil check needs to happen before you read from x.UserLoginAttempts.

auth.go
```diff
@@ -216,12 +245,18 @@ const (
 	AuditActionDeleted                        = "Deleted"
 	AuditActionResetPassword                  = "Reset Password"
 	AuditActionFailedLogin                    = "Failed Login"
+	AuditActionUnlocked                       = "Unlocked"
+	AuditActionlocked                         = "Locked"
```
bmharper: locked -> Locked

auth.go
```diff
@@ -243,12 +278,15 @@ type Central struct {
...
 	PasswordExpiresAfter   time.Duration
+	MaxFailedLoginAttempts int // only applies if EnableAccountLocking is true
+	EnableAccountLocking   bool
 	userStoreMergeTicker   *time.Ticker
 	mergeCount             int
 	ldapUsed               bool
 	ldapMergeTickerSeconds time.Duration
 	tickerStopReq          chan bool
 	tickerStopResp         chan bool
+	LockingPolicy          LockingPolicy
```
bmharper: Follow the convention of the local code, and place public members before private

auth.go
```diff
@@ -503,6 +544,19 @@ func (x *Central) authenticate(identity, password string) (AuthUser, error) {
 		authTypeCheck = AuthCheckDefault
 	}
 
+	invalidPasswords := x.Stats.UserLoginAttempts[identity]
```
bmharper: In other places in the code, you check first whether UserLoginAttempts is nil, but here you assume it is not. Either always check, or always initialize the map, so that we never have to worry whether it's null or not (I would do the latter)

auth.go
```diff
@@ -512,12 +566,34 @@ func (x *Central) authenticate(identity, password string) (AuthUser, error) {
...
-		return user, x.userStore.Authenticate(identity, password, authTypeCheck)
+		authErr = x.userStore.Authenticate(identity, password, authTypeCheck)
+	}
+
+	if authErr == ErrInvalidPassword {
+		username := user.Username
+		x.Stats.IncrementInvalidPasswords(x.Log)
+		x.Stats.IncrementInvalidPasswordHistory(x.Log, username, clientIPAddress)
+		if x.EnableAccountLocking {
+			if isAdmin, groupErr := x.LockingPolicy.IsLockable(identity); groupErr == nil {
```
bmharper: isAdmin is no longer the right word, nor is groupErr

all_test.go
```diff
@@ -325,6 +324,20 @@ func isPrefix(prefix, str string) bool {
 	return strings.Index(str, prefix) == 0
 }
 
+type TestNonLockableUser struct {
+}
+
+func (x TestNonLockableUser) IsLockable(identity string) (bool, error) {
+	return true, nil
```
bmharper: Why does IsLockable return true, when the user is NOT lockable?
A function's return value must be true to its name.

all_test.go
```diff
@@ -1424,3 +1437,60 @@ func TestAuthArchiveIdentity(t *testing.T) {
...
+
+// Implemented in a smilar way to TestAuthUpdateDuplicateIdentity
+func TestAccountLocking(t *testing.T) {
+	c := setup(t)
+	defer Teardown(c)
+	c.MaxFailedLoginAttempts = 3
+	c.EnableAccountLocking = true
+
+	if isBackendPostgresTest() {
+		user, userErr := c.GetUserFromIdentity(joeEmail)
```
bmharper: This test is not clear enough. I'd expect to see a loop that iterates 4 times, trying to login, and receiving a InvalidPassword error the first 3 time, an then on the 4th time, getting back an AccountLocked error.
And then, on the outside of the loop, I'd expect another loop that switches between Lockable and NonLockable user policy. Basically, the test should be shorter and more concise, and more obvious.

auth.go
```diff
@@ -166,6 +167,8 @@ type CentralStats struct {
 	GoodOnceOffAuth    uint64
 	GoodLogin          uint64
 	Logout             uint64
+	UserLoginAttempts  map[string]uint64
+	loginLock          sync.Mutex
```
bmharper: Make the name of this mutex precisely match the variable that it guards, for example,
`userLoginAttemptsLock`, and then it's a good idea also to comment on it "Guards all access to UserLoginAttempts"

auth.go
```diff
@@ -243,12 +278,15 @@ type Central struct {
 	NewSessionExpiresAfter time.Duration
 	DisablePasswordReuse   bool
 	PasswordExpiresAfter   time.Duration
+	MaxFailedLoginAttempts int // only applies if EnableAccountLocking is true
```
bmharper: Can't we move MaxFailedLoginAttempts out into the LockingPolicy?
And perhaps same for EnableAccountLocking?

db.go
```diff
@@ -82,6 +82,8 @@ type UserStore interface {
 	// TODO RenameIdentity was deprecated in May 2016, replaced by UpdateIdentity. We need to remove this once PCS team has made the necessary updates
 	RenameIdentity(oldIdent, newIdent string) error                        // Rename an identity. Returns ErrIdentityAuthNotFound if oldIdent does not exist. Returns ErrIdentityExists if newIdent already exists.
 	GetUserFromIdentity(identity string) (AuthUser, error)                 // Gets the user object from the identity supplied
+	LockAccount(userId UserId) error                                       // Locks account after a certain number failed log in attempts, determined by MaxFailedLoginAttempts in config
```
bmharper: Comment should probably change - especially if we make the LockingPolicy more flexible.
It's not necessarily after X number of failed login attempts. It could be for any reason.

db.go
bmharper: That looks right. And if you make these suggested changes, then the unit test will probably work regardless of whether it's postgres_backend or not

db_sql.go
```diff
@@ -58,6 +58,11 @@ func (x *sqlUserStoreDB) Authenticate(identity, password string, authTypeCheck A
 		return ErrIdentityAuthNotFound
 	}
 
+	var accountLocked bool
+	if err := x.db.QueryRow(`SELECT coalesce(accountlocked, false) FROM authuserpwd WHERE userid = $1`, userId).Scan(&accountLocked); err != nil || accountLocked {
```
bmharper: You're doing the exact same query as below, just with different fields returned. Just combine them into one query.

all_test.go
```diff
@@ -1424,3 +1441,40 @@ func TestAuthArchiveIdentity(t *testing.T) {
...
+	c.loginDelayFactor = 0 // add 0 ms per invalid login attempt
+	c.Auditor = c
+
+	user, userErr := c.GetUserFromIdentity(joeEmail)
+	if userErr != nil {
+		t.Fatal("Couldn't find setup user")
+	}
+
+	loginAttempts := 7
+	for i := 0; i < 2; i++ {
```
bmharper: This is the correct structure that I was referring to, but just some small changes here would make the code more readable:
1. Don't call it `i`. Instead, call it `lockPass`, as in multiple passes over the same concept.
2. In the outer loop, have two branches in the if statement:

```go
if lockPass ==0 {
    // Enable locking
} else {
    // Disable locking, and assume that the previous pass left the user in a locked state
}
```

all_test.go
```diff
@@ -325,6 +324,24 @@ func isPrefix(prefix, str string) bool {
 	return strings.Index(str, prefix) == 0
 }
 
+type TestNonLockableUser struct {
```
bmharper: It would be a bit less code if you just had one thing like this:
```go
type TestLockableUser struct {
    enableLocking bool
}

func (x TestLockableUser) IsLockable(identity string) (bool, error) {
	return x.enableLocking, nil
}

```

all_test.go
```diff
@@ -1424,3 +1441,40 @@ func TestAuthArchiveIdentity(t *testing.T) {
...
+	c.EnableAccountLocking = true
+	c.loginDelayFactor = 0 // add 0 ms per invalid login attempt
+	c.Auditor = c
+
+	user, userErr := c.GetUserFromIdentity(joeEmail)
+	if userErr != nil {
+		t.Fatal("Couldn't find setup user")
+	}
+
+	loginAttempts := 7
```
bmharper: Probably just make this loginAttempts := x.MaxFailedLoginAttempts * 2

auth.go
```diff
@@ -216,12 +239,18 @@ const (
...
 	AuditActionFailedLogin                    = "Failed Login"
+	AuditActionUnlocked                       = "Unlocked"
+	AuditActionLocked                         = "Locked"
 )
 
 type Auditor interface {
 	AuditUserAction(identity, item, context string, auditActionType AuditActionType)
 }
 
+type LockingPolicy interface {
```
bmharper: There should be a comment here saying:
```go
// LockingPolicy controls on a per-user basis, whether that user's account is automatically
// locked, after a set number of failed login attempts. It was created to disable the locking
// of special accounts, such as administrators or internal infrastructure accounts.
// This only applies if EnableAccountLocking is true
```

auth.go
```diff
@@ -503,6 +543,19 @@ func (x *Central) authenticate(identity, password string) (AuthUser, error) {
 		authTypeCheck = AuthCheckDefault
 	}
 
+	invalidPasswords := x.Stats.UserLoginAttempts[identity]
```
bmharper: You need to take the mutex here.

auth.go
```diff
@@ -512,12 +565,34 @@ func (x *Central) authenticate(identity, password string) (AuthUser, error) {
...
+		authErr = x.userStore.Authenticate(identity, password, authTypeCheck)
 	}
+
+	if authErr == ErrInvalidPassword {
+		username := user.Username
+		x.Stats.IncrementInvalidPasswords(x.Log)
+		x.Stats.IncrementInvalidPasswordHistory(x.Log, username, clientIPAddress)
+		if x.EnableAccountLocking {
+			if isLockable, lockabilityErr := x.LockingPolicy.IsLockable(identity); lockabilityErr == nil {
+				if int(x.Stats.UserLoginAttempts[username]) >= x.MaxFailedLoginAttempts && isLockable {
```
bmharper: Need to be holding the mutex

auth.go
```diff
@@ -794,6 +869,18 @@ func (x *Central) UpdateIdentity(user *AuthUser) error {
...
 
+// Unlock user in the AuthUserStore.
+func (x *Central) UnlockAccount(userId UserId) error {
+	e := x.userStore.UnlockAccount(userId)
+	if e != nil {
+		x.Log.Warnf("Failed to unlock user (%v) (%v)", userId, e)
+		return e
+	}
+
+	x.Log.Infof("Successfully unlocked user (%v)", userId)
```
bmharper: Generally, we just say "Unlocked user", ie don't say "successfully", because every log message, unless it's a failure, is implicitly success.

db.go
```diff
@@ -197,6 +201,9 @@ func (x *dummyUserStore) Authenticate(identity, password string, authTypeCheck A
 	if user == nil {
 		return ErrIdentityAuthNotFound
 	}
+	if user.accountLocked == true {
```
bmharper: No need to add the ` == true`. Since Go is strongly typed, it is 100% unambiguous that an `if` statement must be written against a bool.

db.go
```diff
@@ -330,20 +337,38 @@ func (x *dummyUserStore) GetIdentities(getIdentitiesFlag GetIdentitiesFlag) ([]A
...
 		}
-		list = append(list, AuthUser{v.userId, v.email, v.username, v.firstname, v.lastname, v.mobilenumber, v.telephonenumber, v.remarks, v.created, v.createdby, v.modified, v.modifiedby, v.authUserType, v.archived, v.passwordModifiedDate})
+		list = append(list, AuthUser{v.userId, v.email, v.username, v.firstname, v.lastname, v.mobilenumber, v.telephonenumber, v.remarks, v.created, v.createdby, v.modified, v.modifiedby, v.authUserType, v.archived, v.passwordModifiedDate, v.accountLocked})
 	}
 	return list, nil
 }
 
+func (x *dummyUserStore) LockAccount(userId UserId) error {
+	x.usersLock.RLock()
+	defer x.usersLock.RUnlock()
```
bmharper: This is a ReadWrite lock, but you're modifying the value here, so you need to take the full write-able version of the lock.
Read about the semantics of a ReadWrite lock.

db.go
```diff
@@ -330,20 +337,38 @@ func (x *dummyUserStore) GetIdentities(getIdentitiesFlag GetIdentitiesFlag) ([]A
...
+	if user, exists := x.users[userId]; exists && !user.archived {
+		user.accountLocked = true
+	}
+	return nil
+}
+
+func (x *dummyUserStore) UnlockAccount(userId UserId) error {
+	x.usersLock.RLock()
```
bmharper: Same thing applies about the ReadWrite the lock here.

auth.go
bmharper: Rather do the mutex lock/unlock as above, where you hold it for as short as possible. In this formulation, you're holding the mutex while you do LockAccount, and AuditUserAction. Both of those calls could potentially take a LONG TIME. During that time, nobody else can login. So our "feature" has turned into a "bug".

PR #49, opened 2021-04-20
INF-1305 export user groups json

nkosimarz:
Get raw groups from db for export groups (and users in groups) to json

### DIFF
--- a/roledb.go
+++ b/roledb.go
@@ -352,13 +352,11 @@ func (x *sqlGroupDB) GetGroupsRaw() ([]RawAuthGroup, error) {
 
 	var groups []RawAuthGroup
 	for rows.Next() {
-		name := ""
-		permList := ""
-		var id GroupIDU32
-		if errScan := rows.Scan(&id, &name, &permList); errScan != nil {
+		r := RawAuthGroup{}
+		if errScan := rows.Scan(&r.ID, &r.Name, &r.PermList); errScan != nil {
 			return nil, errScan
 		}
-		groups = append(groups, RawAuthGroup{ID: id, Name: name, PermList: permList})
+		groups = append(groups, r)
 	}
 	return groups, nil
 }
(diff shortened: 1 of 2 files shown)


roledb.go
```diff
@@ -335,6 +342,27 @@ func (x *sqlGroupDB) GetGroups() ([]*AuthGroup, error) {
...
+
+	var groups []RawAuthGroup
+	for rows.Next() {
+		name := ""
+		permList := ""
+		var id GroupIDU32
+		if errScan := rows.Scan(&id, &name, &permList); errScan != nil {
+			return nil, errScan
+		}
+		groups = append(groups, RawAuthGroup{ID: id, Name: name, PermList: permList})
```
FritzOnFire, line 361: So for this you can just create the `RawAuthGroup` and then use it for the scan. Like
```go
r = RawAuthGroup{}
Scan(&r.id, &r.name, r.permList)
```
Makes things a bit simpler.

roledb.go
```diff
@@ -408,6 +436,17 @@ type RoleGroupCache struct {
...
 
+// GetGroupsRaw's results are not cached. The point is to get the current state of the db for the export
+func (x *RoleGroupCache) GetGroupsRaw() ([]RawAuthGroup, error) {
+	x.groupsLock.RLock()
+	groups, err := x.backend.GetGroupsRaw()
+	if err != nil {
+		return nil, err
+	}
+	x.groupsLock.Unlock()
+	return groups, nil
```
FritzOnFire, line 447: As discussed, this should just be the one liner returning the result of the function.

FritzOnFire (changes requested):
Some minor changes, but LGTM

FritzOnFire (approved):
LGTM

jasonkofo (approved):

jasonkofo (approved):

PR #51, opened 2021-06-07
Fix incorrect "Modified By" field on user permissions config page

bmharper:
This is a fix for https://imqssoftware.atlassian.net/browse/INF-1369

The two commits contain detailed comments. I have tested this by manually fudging my DB, but haven't tested it with a real LDAP integration.

### DIFF
--- a/all_test.go
+++ b/all_test.go
@@ -868,10 +868,10 @@ func TestAuthResetPassword(t *testing.T) {
 	c := setup(t)
 	defer Teardown(c)
 
-	if _, err := c.ResetPasswordStart(notFoundUserId, time.Now()); err != ErrIdentityAuthNotFound {
+	if _, err := c.ResetPasswordStart(notFoundUserId, time.Now()); !errors.Is(err, ErrIdentityAuthNotFound) {
 		t.Fatalf("ResetPasswordStart should fail with ErrIdentityAuthNotFound instead of %v", err)
 	}
-	if err := c.ResetPasswordFinish(joeUserId, "", "12345"); err != ErrInvalidPasswordToken {
+	if err := c.ResetPasswordFinish(joeUserId, "", "12345"); !errors.Is(err, ErrInvalidPasswordToken) {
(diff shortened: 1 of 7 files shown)


all_test.go
```diff
@@ -1300,8 +1300,8 @@ func TestAuthUpdateIdentity(t *testing.T) {
 		ModifiedBy:      newModifiedBy,
 		Type:            UserTypeLDAP,
 	}
-	if err := c.UpdateIdentity(&notFoundUser); err != ErrIdentityAuthNotFound {
-		t.Fatalf("TestUpdateIdentity failed: Expected ErrIdentityAuthNotFound, but got: %v", err)
+	if err := c.UpdateIdentity(&notFoundUser); err == nil {
+		t.Fatalf("Expected ErrIdentityAuthNotFound, but got: %v", err)
```
nkosimarz, line 1304: wont the err in %v be nil always? shouldn't it be !errors.Is(err, ErrIdentityAuthNotFound)?

all_test.go
```diff
@@ -1430,8 +1430,8 @@ func TestAuthArchiveIdentity(t *testing.T) {
 	// Using an ldap backend, authentication will succeed, as we cannot archive users on the ldap system.
 	if !*backend_ldap {
 		// Try to authenticate with archived user
-		if _, err := c.GetTokenFromIdentityPassword(joeEmail, joePwd); err != ErrIdentityAuthNotFound {
-			t.Fatalf("TestArchiveIdentity failed, archived user should not be allowed to authenticate: %v", err)
+		if _, err := c.GetTokenFromIdentityPassword(joeEmail, joePwd); err == nil {
+			t.Fatalf("Archived user should not be allowed to authenticate: %v", err)
```
nkosimarz, line 1434: is the %v needed since its nil? same for the rest below

nkosimarz (changes requested):

all_test.go
bmharper: I just thought I was being a good Samaritan, cleaning up Stewart's old cruft.. but sure.. you're not wrong!

PR #52, opened 2021-07-27
Change GroupIDsToNames to take a 'cache' parameter to avoid common issue

bmharper:
The issue is that GroupIDstoNames often ends up being called in a loop
over many users, and every invocation ends up hitting the DB. By forcing
the user to pass a cache parameter through, we quickly reduce DB fetches.


bosvos (approved):
LGTM
This will result in a new version of authhaus, correct?

PR #56, opened 2022-06-27
ASG-1882 : Add username update to MSAAD synch

bosvos:
Correct an oversight in the MSAAD synchronisation process where the username was not populated, neither with the email (as in the OAuth flow) nor with the UserPricipalName.


FritzOnFire (approved):
LGTM

PR #58, opened 2022-09-01
BAZ-202 : Fix passthrough auth session bug.

bosvos:
Passthrough auth never created oauth sessions since it relied on entries in oauthchallenge. It appeared to work for a while after login because there were normal sessions created in authsession, which quickly got expired on the next token refresh from MSAAD (no matching oauthsession).


LeonLeibbrandt (approved):
LGTM

PR #59, opened 2023-02-28
Add authuserstore db triggers. 

JeanNicholson:
fix: Add authuserstore db triggers. 

Add triggers to ensure that created, createdby and modified for the authuserstore table gets set correctly. 

- created - this gets set to NOW() when a record gets inserted. Once the record has been inserted this value should not change and the update trigger ensures this.
- createdby - if this value is null when a record gets inserted an exception will be raised. Once the record has been inserted this value should not change and the update trigger ensures this.
- modified - this value should always be set to NOW() whenever a record gets inserted or updated.
		
The naming convention used for the functions and triggers is as follows:
{database name}_{schema name}_{table name}_{type of trigger}

type of trigger:
- first character denoted when the action happens. 'b' for before, 'a' for after.
- second character denoted the type of action. 'i' for insert, 'u' for update.
- the last character just indicates that this is for a trigger and 't' is used.

Jira Ticket: ASG-2210

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,0 +1,19 @@
+# Changelog
+
+## current
+
+## v1.0.27
+
+* fix: Add authuserstore db triggers. (ASG-2210)
+
+## v1.0.26
+
+* fix: passthrough auth session bug. (BAZ-202)
+
+## v1.0.25
+
+* Modified the MSAAD synchronisation process to set blank usernames for users in Auth to the email provided from AD.
+ 
+## v1.0.24
+
+* Deprecated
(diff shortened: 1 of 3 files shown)


bosvos:
LGTM - except:
1. Readme file should be updated with new "requirements" e.g. _must_ supply createdby...
2. Even though we have the git log...we should maintain an explicity change.log file with more detailed information per version...you don't have to backdate - just add the new stuff. Look at git@github.com/IMQS/www for an example.

bosvos (approved):

PR #62, opened 2023-06-26
ASG-2622: Change all debug and trace logs to info

bosvos:


msaad.go
```diff
@@ -762,9 +768,9 @@ func (m *MSAAD) populateAADRoles(users []*msaadUser) error {
 						break
 					}
 					if m.parent.MSAAD.Config.Verbose {
-						m.parent.Log.Debugf("User %v (%v): %v\n", user.profile.bestEmail(), user.profile.ID, j)
+						m.parent.Log.Infof("User %v (%v): %v\n", user.profile.bestEmail(), user.profile.ID, j)
 						for _, u := range j.Value {
-							m.parent.Log.Debugf("%v MSAAD User Permission: (%v)", user.profile.bestEmail(), u)
+							m.parent.Log.Infof("%v MSAAD User Permission: (%v)", user.profile.bestEmail(), u)
```
CelesteNaude, line 773: The way these logs are written feel more as if its for debugging. They don't provide anything on an info level.

bosvos: Sure - the issue is that Auth does not currently provide for setting the log level - it's all default **Info**. Hence my issue with the pattern - you'll see an **if** testing "Verbose" a bit higher up which covers a few lines there, since all of it is, as you rightly pointed out, debug code.

anzelIMQS (approved):
LGTM

CelesteNaude (approved):
LGTM

PR #64, opened 2023-09-06
Asg 1990 jv missing auth groups

bosvos:


FritzOnFire (commented):
Please also add an extra check before https://github.com/IMQS/authaus/blob/master/msaad.go#L592 to ensure that `roleGroups.idToGroup[id]` actually exists.

roledb.go
```diff
@@ -293,19 +294,38 @@ func ReadRawGroups(importedGroups []RawAuthGroup) ([]AuthGroup, error) {
...
-			names[i] = cache[gid]
+			//names[i] = cache[gid]
+			names = append(names, cache[gid])
 		} else {
-			if group, err := db.GetByID(gid); err != nil {
+			// why is this not retrieved in one go? the below would result in 'n' nr of individual queries to db
+			// the below should only happen at most once per request
+			// if no new groups are added before the next request, cache should be up to date
+			// CAVEAT - In case of missing group, the cache will be retrieved EVERY time the group is referenced.
+			// Consider using a 'populated' flag
```
FritzOnFire, line 310: We should rather create a function to allow people to clean up permits on users

roledb.go
```diff
@@ -293,19 +294,38 @@ func ReadRawGroups(importedGroups []RawAuthGroup) ([]AuthGroup, error) {
...
+					if cache[giddb.ID] == "" {
+						cache[giddb.ID] = giddb.Name
+					}
+				}
+				// cycle through new groups, find in cache (based on new group id)
+				if cache[gid] != "" {
+					//names[i] = cache[gid]
+					names = append(names, cache[gid])
+				}
 			}
```
FritzOnFire, line 325: Had a long rant here about how we should keep the old code, but based on your comment it looks like we were both wrong as to how this function is *used*.

The three places that we call this function in *our* auth service, we feed it an empty cache... which means that we will always end up hitting the db ALOT. My recommendation would be that the calling function should initialize the cache, to allow this function to remain smaller. But that would require us to create a function that does
```go
			cache := map[GroupIDU32]string{}

			if groupsDB, err := db.GetGroups(); err != nil {
				return nil, err
			}

			// populate groups cache
			for _, giddb := range groupsDB {
				if cache[giddb.ID] == "" {
						cache[giddb.ID] = giddb.Name
				}
			}
			return cache, nil
```

I would also compromise with checking the `len` of `cache` at the start of the function and then calling the new function to create a cache of the groups.

roledb.go
```diff
@@ -293,19 +294,38 @@ func ReadRawGroups(importedGroups []RawAuthGroup) ([]AuthGroup, error) {
 // The 'cache' parameter is used to speed up subsequent calls to this function,
 // because this function tends to get used in loops.
 func GroupIDsToNames(groups []GroupIDU32, db RoleGroupDB, cache map[GroupIDU32]string) ([]string, error) {
-	names := make([]string, len(groups))
-	for i, gid := range groups {
+	// A group in 'groups' may not exist in the RoleGroupDB and we don't want to return a blank item in the string array.
```
FritzOnFire, line 297: This comment only makes sens if someone sees the previous code. I don't think we need it.

roledb.go
bosvos: Sure. I would suggest we add it to the other tickets in the epic.

roledb.go
```diff
@@ -245,7 +245,8 @@ func PermitResolveToList(permit []byte, db RoleGroupDB) (PermissionList, error)
 	if groupIDs, err := DecodePermit(permit); err == nil {
 		for _, gid := range groupIDs {
 			if group, egroup := db.GetByID(gid); egroup != nil {
-				return nil, egroup
+				// this will error out if the group does not exist - we should log and continue
+				//return nil, egroup
```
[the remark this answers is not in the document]
bosvos: Forgot about this comment in the code. Was meaning to do what you suggest.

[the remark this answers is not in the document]
bosvos: Actually - i've made another plan there. I don't want to break the current expectation from the calling code which lives in imqsauth. Will submit soon.

roledb.go
```diff
@@ -293,19 +294,38 @@ func ReadRawGroups(importedGroups []RawAuthGroup) ([]AuthGroup, error) {
 // The 'cache' parameter is used to speed up subsequent calls to this function,
 // because this function tends to get used in loops.
 func GroupIDsToNames(groups []GroupIDU32, db RoleGroupDB, cache map[GroupIDU32]string) ([]string, error) {
-	names := make([]string, len(groups))
-	for i, gid := range groups {
+	// A group in 'groups' may not exist in the RoleGroupDB and we don't want to return a blank item in the string array.
+	// Instead we use append. It does mean a new slice variable gets created, but the overhead is minimal.
+	// Since the capacity of the slice is already maxed, no new underlying array is allocated during append.
```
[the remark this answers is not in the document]
bosvos: Can you elaborate?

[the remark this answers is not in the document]
FritzOnFire: The first line talks about the down sides of using append, but they are not relevant in this function, as we pre-allocate enough for re-alloc to never occur, so the comment has no relation to anything in this function.

Then the second comment confirms the understanding of what I just commented, meaning that the second comment invalidates the statement made in the first comment.

But the second comment it self is written as if to be a response on the first comment. But both were added in this PR... so if we don't state the first comment, then there is no need to answer it, and the code can just be what it is.

If you really want, you can leave a comment here like `Pre-allocate slice to improve performance`. But I personally feel like there is no other reason to pre-allocate, so the code kinda speaks for it self already.

roledb.go
```diff
@@ -289,23 +294,43 @@ func ReadRawGroups(importedGroups []RawAuthGroup) ([]AuthGroup, error) {
...
+				for _, giddb := range groupsDB {
+					if cache[giddb.ID] == "" {
+						cache[giddb.ID] = giddb.Name
+					}
+				}
+				// cycle through new groups, find in cache (based on new group id)
+				if cache[gid] != "" {
+					names = append(names, cache[gid])
+				}
 			}
```
FritzOnFire, line 330: I really don't like all of this logic living in an `else` statement... Most of the lines in this function now have nothing to do with the function... Please create the function I asked for, even if you have to pass in the `cache` as a parameter to the function.

Also, we still need to return an error, you can take the same approach as you did with `PermitResolveToList` and just keep the error around to return at the very end. You will most likely have to write the error message your self.

As to your comment, I would remove the first 3 lines, as it only makes sens if the developer was also looking at the old code. But the last two lines I feel is important in case we ever have performance issues and someone needs to investigate.

FritzOnFire (changes requested):

roledb.go
bosvos: I think between this comment and the previous one on the older code we've missed each other. Did you mean "the calling function should _populate_ the cache" instead? Since the calling function _does_ initialise the cache...
And then - would you prefer we keep the old code and provide a _new_ function as an enhancement (and replace some of the new code in the new function with another function so as to simplify the logic there)?
I'll do what i think you want and perhaps discuss offline if still required.

FritzOnFire (approved):
So I don't know what the precedence is about returning a `nil` instead of an array, but I am pretty sure that we can put anything in there as along as we also return an error, so I think this code is good.

Please just remove the stray logging left over in the unit tests.

PR #66, opened 2023-11-08
Added exempt from expiring functionality. 

JeanNicholson:
Jira Ticket: ASG-3055


anzelIMQS (approved):
LGTM

FritzOnFire:
@CelesteNaude so this is a public repo that can be used by any one in any piece of software. We should not have anything in here that is specifically for our company.

JeanNicholson:
> I understand the functionality of this, but can't help feeling a bit uneased that we have an option to give auth a random list of users to skip session expiry (although not having session expiry for our type of product is also weird). Can the list of users that we exempt be bespoke (like `imqs_importer`)? Or am I overthinking it?

Maybe the variable names can be a bit more clear, could consider changing it perhaps. This is not to change session expiry but rather to change what users login/password expires and areas where that effect session expiry. I think your comment might still be valid in that case, but I don't think we need to worry about that. Main reason is that we have much "bigger" things controlled by config for example just the fact that sessions expiry/password expiry is active or not.

JeanNicholson:
> @CelesteNaude so this is a public repo that can be used by any one in any piece of software. We should not have anything in here that is specifically for our company.

Ah also a very valid point. Forgot about that.

CelesteNaude (approved):
Thanks for the feedback. LGTM

db_sql.go
```diff
@@ -97,20 +98,34 @@ func (user *sqlUser) toAuthUser() *AuthUser {
...
+		if user == username {
+			return true
+		}
+	}
+	return false
+}
+
 func (x *sqlUserStoreDB) Authenticate(identity, password string, authTypeCheck AuthCheck) error {
-	row := x.db.QueryRow(`SELECT userid FROM authuserstore WHERE (LOWER(email) = $1 OR LOWER(username) = $1) AND (archived = false OR archived IS NULL)`, CanonicalizeIdentity(identity))
+	row := x.db.QueryRow(`SELECT userid, userName FROM authuserstore WHERE (LOWER(email) = $1 OR LOWER(username) = $1) AND (archived = false OR archived IS NULL)`, CanonicalizeIdentity(identity))
```
FritzOnFire, line 125:
```suggestion
	row := x.db.QueryRow(`SELECT userid, username FROM authuserstore WHERE (LOWER(email) = $1 OR LOWER(username) = $1) AND (archived = false OR archived IS NULL)`, CanonicalizeIdentity(identity))
```

db_sql.go
```diff
@@ -97,20 +98,34 @@ func (user *sqlUser) toAuthUser() *AuthUser {
...
 	var userId int64
-	if err := row.Scan(&userId); err != nil {
+	var userName string
```
FritzOnFire, line 127:
```suggestion
	var username string
```

db_sql.go
```diff
@@ -97,20 +98,34 @@ func (user *sqlUser) toAuthUser() *AuthUser {
...
+	if err := row.Scan(&userId, &userName); err != nil {
```
FritzOnFire, line 128:
```suggestion
	if err := row.Scan(&userId, &username); err != nil {
```

db_sql.go
```diff
@@ -136,7 +151,7 @@ func (x *sqlUserStoreDB) Authenticate(identity, password string, authTypeCheck A
 		return ErrInvalidPassword
 	}
 
-	if x.passwordExpiry != 0 && authTypeCheck&AuthCheckPasswordExpired != 0 {
+	if x.passwordExpiry != 0 && authTypeCheck&AuthCheckPasswordExpired != 0 && !x.ExemptFromExpiryCheck(userName) {
```
FritzOnFire, line 154:
```suggestion
	if x.passwordExpiry != 0 && authTypeCheck&AuthCheckPasswordExpired != 0 && !x.ExemptFromExpiryCheck(username) {
```

FritzOnFire (changes requested):
One small niggly thing... and then there is the unit test issue... because we don't have a unit test for `PasswordExpiresAfter`, it feels bad asking for a unit test for `UsersExemptFromExpiring`.

Would you mind just adding two TODO's after the TODO in `all_test.go`. One that just mentions that we need to add a unit test for `PasswordExpiresAfter` and one for `UsersExemptFromExpiring`

DevinFledermaus (approved):
LGTM

all_test.go
```diff
@@ -23,6 +23,10 @@ TODO: Add test that verifies that SetPassword does not create an identity if tha
 Create a test Postgres database:
 	create database authaus_test owner = authaus_test;
 
+Create a test for password expiry (PasswordExpiresAfter).
+
+Create a test for users exempt from expiry (UsersExemptFromExpiring).
```
FritzOnFire, line 28: So `Create a test Postgres database` is an instruction :P not a TODO. Please move these two comments to below the TODO, and give them each their own TODO, and reword to `Add test` instead of `Create a test`

FritzOnFire (approved):
One small change, about some comments, and when updating the `changelog.md` please remember to add a `!` as this is a breaking change.

And, in case you have not encounter this, even thou this is a breaking change, we can not bump the major version as go-lang has some annoying rules around creating a new major version of your library... so this will just be a minor version bump :/

PR #67, opened 2023-11-15
ASG-2921: New functions on db to support retrieving tokens.

bosvos:


auth.go
```diff
@@ -996,7 +1004,54 @@ func (x *Central) GetAuthenticatorIdentities(getIdentitiesFlag GetIdentitiesFlag
 func (x *Central) GetRoleGroupDB() RoleGroupDB {
 	return x.roleGroupDB
 }
+func RemoveGroupFromAllUsers(c *Central, groupIDString string) error {
```
FritzOnFire, line 1007:
```suggestion
func (c *Central) RemoveGroupFromAllUsers(groupIDString string) error {
```

FritzOnFire, line 1007: Please leave an empty line between functions

auth.go
```diff
@@ -996,7 +1004,54 @@ func (x *Central) GetAuthenticatorIdentities(getIdentitiesFlag GetIdentitiesFlag
...
+
+		// Encode the permit
+		if modified {
+			perm.Roles = EncodePermit(groups)
+			fmt.Printf("Removing group %v: setting permit for %v\n", groupID, user.Email)
+
+			//Set the permit
+			err = c.SetPermit(user.UserId, perm)
+			if err != nil {
+				fmt.Printf("Error: %v\n", err)
```
FritzOnFire, line 1047: I think we should remove this line, as the caller of this function can log this if it feels like it.

db.go
```diff
@@ -381,6 +383,15 @@ func (x *dummySessionDB) Read(sessionkey string) (*Token, error) {
 	return token, nil
 }
 
+func (x *dummySessionDB) GetAllTokens(includeExpired bool) ([]*Token, error) {
+	//TODO implement me
+	panic("implement me")
+}
+func (x *dummySessionDB) GetAllOAuthTokenIDs() ([]string, error) {
```
FritzOnFire, line 390: Please leave an empty line between functions.

auth.go
```diff
@@ -996,7 +1004,54 @@ func (x *Central) GetAuthenticatorIdentities(getIdentitiesFlag GetIdentitiesFlag
...
+			if g == groupID {
+				groups = append(groups[:i], groups[i+1:]...)
+				modified = true
+			}
+		}
+
+		// Encode the permit
+		if modified {
+			perm.Roles = EncodePermit(groups)
+			fmt.Printf("Removing group %v: setting permit for %v\n", groupID, user.Email)
```
FritzOnFire, line 1042: I would rather make use of `c.Log.Debugf`. But Infof might also be fine, as I don't see something like this filling up a log file, or standard out.

auth.go
```diff
@@ -996,7 +1004,54 @@ func (x *Central) GetAuthenticatorIdentities(getIdentitiesFlag GetIdentitiesFlag
...
+			if err != nil {
+				fmt.Printf("Error: %v\n", err)
+				return err
+			}
+		}
+	}
+
+	return nil
+}
 func (x *Central) IsShuttingDown() bool {
```
FritzOnFire, line 1055: Please leave an empty space between functions.

db_sql.go
```diff
@@ -648,28 +649,97 @@ func (x *sqlSessionDB) Read(sessionkey string) (*Token, error) {
...
+		} else {
+			token.Identity = user.Email
+		}
+		if oauthID.Valid {
+			token.OAuthSessionID = oauthID.String
+		}
+
+		return token, nil, true
+	}
+	return nil, nil, false
```
FritzOnFire, line 683: If I am understanding this code correctly, then I would remove the `else` and just have it's content be the rest of the function.

db_sql.go
```diff
@@ -648,28 +649,97 @@ func (x *sqlSessionDB) Read(sessionkey string) (*Token, error) {
...
-			token.Username = user.Username
-			token.Email = user.Email
-			if user.Type == UserTypeLDAP {
-				token.Identity = user.Username
+			return nil, nil
+		}
+	}
+}
+
+func (x *sqlSessionDB) populateToken(token *Token, epermit string, userId int64, oauthID sql.NullString) (*Token, error, bool) {
```
FritzOnFire, line 661: I might be wrong... but I don't think you need the `bool` at the end. At least, not with the code that I can see in this PR.

db_sql.go
```diff
@@ -648,28 +649,97 @@ func (x *sqlSessionDB) Read(sessionkey string) (*Token, error) {
 			return nil, err
 		}
 	} else {
-		if err := token.Permit.Deserialize(epermit); err != nil {
-			return nil, err
+		t, err2, done := x.populateToken(token, epermit, userId, oauthID)
```
FritzOnFire, line 652: The current implementation of `populateToken` will always return true for `done`. I think you should rather check the `err`, like the previous code did.

FritzOnFire (changes requested):
Quite a few changes, but it should simplify the code a bit.

auth.go
bosvos: Keeping it as Infof for now. We still look at the auth logfile very regularly to check if groups have been recognized and appropriate action has been taken on the user.

db_sql.go
bosvos: Removed.

PR #68, opened 2024-03-27
Harden MSAAD sync (ASG-3350)

bosvos:
Fix deadlock bugs

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -733,7 +733,14 @@ func (x *Central) StartMergeTicker() error {
 	}
 	x.syncMergeTickerStopRequest = make(chan bool)
 	x.syncMergeTickerStopResponse = make(chan bool)
+
 	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				x.Log.Errorf("MergeTick panic: %v", r)
+			}
+			x.Log.Info("Go routine for merge tick is stopping")
+		}()
 		x.MergeTick()
 		for {
 			select {
@@ -750,6 +757,12 @@ func (x *Central) StartMergeTicker() error {
 }
 
 func (x *Central) MergeTick() {
+	defer func() {
+		if r := recover(); r != nil {
+			x.Log.Errorf("MergeTick panic: %v", r)
+		}
+	}()
+
 	timeStart := time.Now()
 
 	if x.ldap != nil {
--- a/msaad.go
+++ b/msaad.go
@@ -349,6 +349,8 @@ func (m *MSAAD) SynchronizeUsers() error {
 	if len(m.Config.RoleToGroup) != 0 || len(m.Config.DefaultRoles) != 0 {
 		err = m.populateAADRoles(aadUsers)
 		if err != nil {
+			// Quit without merging, because we run the risk of archiving users that we did not successfully receive.
+			// See "if !insideAAD" in the code below.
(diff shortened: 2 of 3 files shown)


msaad.go
```diff
@@ -751,32 +756,58 @@ func numParallelFetchThreads(nItems int) int {
...
 		t := i % nThreads
 		threadGroups[t] = append(threadGroups[t], u)
 	}
 	wg := sync.WaitGroup{}
-	errChan := make(chan error, nThreads)
-
+	var errGlobal error
 	startTime := time.Now()
+	i := 0
 	for _, threadGroupOuter := range threadGroups {
```
FritzOnFire, line 774: Pretty sure you can just piggy back off this.
```suggestion
	for i, threadGroupOuter := range threadGroups {
```

msaad.go
```diff
@@ -870,7 +899,9 @@ func (m *MSAAD) doHTTP(request *http.Request) (*http.Response, error) {
 
 	request.Header.Set("Authorization", "Bearer "+token)
 
-	return http.DefaultClient.Do(request)
+	client := http.DefaultClient
+	client.Timeout = 5 * time.Second
```
FritzOnFire, line 903: That is a tight deadline, I am going to assume there is a good reason for this.

auth.go
```diff
@@ -733,7 +733,14 @@ func (x *Central) StartMergeTicker() error {
 	}
 	x.syncMergeTickerStopRequest = make(chan bool)
 	x.syncMergeTickerStopResponse = make(chan bool)
+
 	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				x.Log.Errorf("MergeTick panic: %v", r)
```
FritzOnFire, line 740: So in mm/rad we have this whole `GetStack` function that we use when logging the error. Does the `r` have enough info that we don't need to dump the stack?

Here is the `GetStack` function for reference
```go
func GetStack() string {
	buf := make([]byte, 1024)
	for {
		n := runtime.Stack(buf, false)
		if n < len(buf) {
			return string(buf[:n])
		}
		buf = make([]byte, 2*len(buf))
	}
}
```

msaad.go
```diff
@@ -913,6 +946,7 @@ func (m *MSAAD) getBearerToken() (token string, expiresAt time.Time, err error)
 	}
 
 	token = tokenJSON.AccessToken
+	// can time.Duration(string) not PANIC?
```
FritzOnFire, line 949: `tokenJSON.ExpiresIn` is an `int`. Go lang would not allow you to put anything in there that is not an `int`. Golang also sets default values.

FritzOnFire (approved):
Some questions and statements.

msaad.go
bosvos: of course

msaad.go
bosvos: Purely because there was no deadline in place prior...suggestions?

auth.go
bosvos: Good call...will implement and test.

bosvos: Was not enough info initially (only high level error message, e.g. `division by zero`). Will implement GetStack as is...

FritzOnFire (approved):
LGTM

PR #72, opened 2025-03-06
Add enabled and disabled audit user types

JeanNicholson:
Create a special use case for the adding and removing of the ‘enabled’ permission. Add two new action types: Disabled: User Profile and Enabled: User Profile to the audit trail.

refs: NEXUS-4244


FritzOnFire (approved):
LGTM

bosvos (approved):

PR #74, opened 2025-03-28
Add and expose user info diff functions (NEXUS-4317)

bosvos:
Make UserInfoDiff public (NEXUS-4317)
New function UserDiffLogMessage


JeanNicholson (approved):
LGTM

PR #75, opened 2025-03-28
Add new columns under "User List" exports

JeanNicholson:
Add the Termination date, Last Login date, Disabled date and Enabled date to the "User List" export. 

refs: NEXUS-4245

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -1011,6 +1011,19 @@ func (x *Central) UnArchiveIdentity(userId UserId) error {
 	return nil
 }
 
+func (x *Central) SetUserStats(userId UserId, action string) error {
+	e := x.userStore.SetUserStats(userId, action)
+	if e != nil {
+		x.Log.Warnf("SetUserStats failed: (%v), (%v)", userId, e)
+		return e
+	}
+	return nil
+}
+
+func (x *Central) GetUserStats(userId UserId) (userStats, error) {
+	return x.userStore.GetUserStats(userId)
+}
+
 // GetUserFromIdentity gets AuthUser object from either an email address or a username.
 func (x *Central) GetUserFromIdentity(identity string) (AuthUser, error) {
 	user, e := x.userStore.GetUserFromIdentity(identity)
--- a/auth.go
+++ b/auth.go
@@ -1012,10 +1012,8 @@ func (x *Central) UnArchiveIdentity(userId UserId) error {
 }
 
 func (x *Central) SetUserStats(userId UserId, action string) error {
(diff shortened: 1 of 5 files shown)


auth.go
```diff
@@ -1011,6 +1011,19 @@ func (x *Central) UnArchiveIdentity(userId UserId) error {
 	return nil
 }
 
+func (x *Central) SetUserStats(userId UserId, action string) error {
+	e := x.userStore.SetUserStats(userId, action)
+	if e != nil {
+		x.Log.Warnf("SetUserStats failed: (%v), (%v)", userId, e)
```
bosvos, line 1017: I would change this to error. It is important enough and affects our ability to accurately provide audit information.

db_sql.go
```diff
@@ -542,6 +542,64 @@ func (x *sqlUserStoreDB) UnarchiveIdentity(userId UserId) error {
...
+		update, err = tx.Exec(`INSERT INTO authuserstats (user_id, enabled_date, disabled_date)
+							   VALUES ($1, NULL, NOW())
+							   ON CONFLICT (user_id)
+							   DO UPDATE SET enabled_date = NULL, disabled_date = NOW()`, userId)
+	}
+	if err != nil {
+		return fmt.Errorf("could not update auth user stats: %v", err)
+	}
+	if affected, _ := update.RowsAffected(); affected != 1 {
+		return fmt.Errorf("user stats could not be updated: %v", ErrIdentityAuthNotFound)
```
bosvos, line 577: This error is not necessarily because the user was not found...it could be because rowsaffected returned more than one row...even though the sql makes that (hopefully) impossible, we should probably make it clearer.

db_sql.go
```diff
@@ -542,6 +542,64 @@ func (x *sqlUserStoreDB) UnarchiveIdentity(userId UserId) error {
...
+							   ON CONFLICT (user_id) 
+							   DO UPDATE SET enabled_date = NOW(), disabled_date = NULL`, userId)
+	}
+	if action == UserStatActionDisable {
+		update, err = tx.Exec(`INSERT INTO authuserstats (user_id, enabled_date, disabled_date)
+							   VALUES ($1, NULL, NOW())
+							   ON CONFLICT (user_id)
+							   DO UPDATE SET enabled_date = NULL, disabled_date = NOW()`, userId)
+	}
+	if err != nil {
```
bosvos, line 573: I would add an else with an error in here, so that if we accidentally miss handling a new action type, we'll see it.

bosvos (commented):
A few comments

bosvos (approved):
LGTM

PR #76, opened 2025-08-10
Jv auth check log

bosvos:
Add new feature to authaus to support "auth check" logging for usage analysis. This pull request is only for the 'authaus' repository (being used by 'imqsauth') and should support the in-memory structure and database persistence.

Auth usage tracking requirements
In our auth system, we can verify user session activity by checking the
/check endpoint with a valid session token. This endpoint returns a
JSON object containing the user's email, username, and a list of groups
the user belongs to. This information can be used to track user activity and system usage in general.

Implement a mechanism that will record these 'check' calls for analysis at a later time.

Design
The auth system should implement a mechanism to track user activity by logging
/check endpoint calls in memory, and periodically flushing this data to persistent storage.

It is important that the user identity (e.g. username) as well as the session token
is logged, so that we can correlate the user activity with the session from other
sources (e.g. router).

The in-memory data structure should be a time log of /check requests, with
timestamp, session token and identity.

The data should be flushed to storage at regular intervals (e.g. every 1 minute).
Expired sessions are not to be recorded, since they should fail the initial check in any case.

Implementation
The implementation of the auth usage tracking can be done as follows:

Create an in-memory data structure to hold the user check request logs.
Implement a middleware that intercepts the /check endpoint calls and logs
the user identity and session token along with the timestamp.
Ensure that the logs are stored in a format that allows for easy querying
Set up a periodic task that flushes the in-memory logs to the database
and analysis of user activity.
4.1 Assume a persistence abstraction from authaus and code against that.
4.2 Clear the in-memory logs after flushing to the database.
Add a configuration setting to enable or disable the auth usage tracking feature.
Implement error handling to ensure that any issues with logging or database
operations do not disrupt the user experience.


permitchecklog.go
```diff
@@ -0,0 +1,187 @@
...
+		} else {
+			// Only clear logs after successful persistence
+			// Check if new logs were added during persistence
+			if len(t.logs) >= logsCount {
+				// Remove the persisted logs (first logsCount entries)
+				t.logs = t.logs[logsCount:]
+			} else {
+				// Shouldn't happen, but clear all if somehow we have fewer logs
+				t.logs = t.logs[:0]
+			}
```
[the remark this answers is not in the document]
bosvos: I don't understand what you mean by "assumes logs are only appended during persistence"...which is the only portion we care about when dealing with the clearing of the logs...after clearing "up to" the persistence count, the persistence is done and t.logs carries on being populated as normal - the goroutine would have exited.

permitchecklog.go
```diff
@@ -0,0 +1,204 @@
...
+	logs             []CheckLogEntry
+	mutex            sync.RWMutex
+	stopChan         chan struct{}
+	doneChan         chan struct{}
+	flushDoneChan    chan struct{}
+	stopOnce         sync.Once
+	flushTicker      *time.Ticker
+	flushing         bool // Track if a flush is in progress
+	db               *sql.DB
+	droppedLogsCount int64
```
LeonLeibbrandt, line 31: Maybe add droppedLogsCount to the database records, since there is no other exposure to it; it is only used in testing?

LeonLeibbrandt (approved):
LGTM

.gitignore
```diff
@@ -1,4 +1,10 @@
...
-test-out/
-/testconf/
+/authaus.out
 /log/
+/testconf/
+test-out/
+/out
+/test.log
+/test-2025-08-07T06-17-16.613.log
+/test-2025-08-07T06-18-32.789.log
```
FritzOnFire: We should never have any `*.log` files in this repo. Please rather change this to
```suggestion
*.log
```

.gitignore
```diff
@@ -1,4 +1,10 @@
...
+/TestAuthRoleDB_MissingGroups_in_github_com_IMQS_authaus.test.exe
```
FritzOnFire: We should never have any `*.exe` files in this repo. Please rather change this to
```suggestion
*.exe
```

config.go
```diff
@@ -166,6 +166,23 @@ type ConfigUserStoreDB struct {
 	UsersExemptFromExpiring []string // List of users that are not subject to password expiry. Username will be used for comparison.
 }
 
+// UsageTrackingConfig controls session check usage tracking
+type ConfigUsageTracking struct {
+	Enabled              bool `json:"enabled"`                // Enable/disable usage tracking
+	FlushIntervalSeconds int  `json:"flush_interval_seconds"` // Flush interval in seconds (default: 60)
+	MaxEntries           int  `json:"max_entries"`            // Maximum number of entries to keep in memory before flushing (default: 1000)
```
FritzOnFire:
```suggestion
	MaxEntries           int  `json:"max_entries"`            // Maximum number of entries to keep in memory before flushing (default: 10 000)
```

permitchecklog.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+}
+
+// LogCheck adds a check request to the in-memory log
+func (t *CheckUsageTracker) LogCheck(sessionToken string, token *Token) {
+	if !t.config.Enabled {
+		return
+	}
+
+	entry := CheckLogEntry{
```
FritzOnFire: Bad bot. We should be handling the pointer to the struct, instead of the struct it self.
```suggestion
	entry := &CheckLogEntry{
```

permitchecklog.go
```diff
@@ -0,0 +1,204 @@
...
+		for {
+			select {
+			case <-t.flushTicker.C:
+				t.flush()
+			case <-t.stopChan:
+				t.flushTicker.Stop()
+				t.flush() // Final flush before stopping
+				return
+			}
+		}
```
FritzOnFire: Bad bot.
* There needs to be an array of `*CheckLogEntry` in this go routine that is pre-allocated with a capacity of `t.config.MaxEntries`.
* In the select we need to be a wait on the log channel that returns a `*CheckLogEntry` that needs to be appended to the array of `*CheckLogEntry`, then the length of the array needs to be compared with `t.config.MaxEntries`. If equal or more we call `t.persistLogs` with the array and we reslice the array to zero length.
* The `case <-t.flushTicker.C:` should be calling `t.persistLogs` and reslice the array to zero length. Please remove the call to the `flush` function.
* The `case <-t.stopChan:` should replace the call to `flush` with `t.persistLogs`

```suggestion
		logs := make(*CheckLogEntry, 0, t.config.MaxEntries)
		for {
			select {
			case entry := <-t.logChan:
				logs = append(logs, entry)
				if len(logs) >= t.config.MaxEntries {
					err := t.persistLogs(logs)
					if err != nil {
						t.log.Errorf("Failed to persist auth check logs: %v", err)
						return
					}
					logs = logs[:0]
				}
			case <-t.flushTicker.C:
				err := t.persistLogs(logs)
				if err != nil {
					t.log.Errorf("Failed to persist auth check logs: %v", err)
					return
				}
				logs = logs[:0]
			case <-t.stopChan:
				t.flushTicker.Stop()
				err := t.persistLogs(logs) // Final flush before stopping
				if err != nil {
					t.log.Errorf("Failed to persist auth check logs: %v", err)
					return
				}
				return
			}
		}
```

@bosvos so when we check the error of `persistLogs`, we MUST bring the service down, as the service failed to do something that it was explicitly configured to do... but I don't really know how to tell Copilot this. We could replace the `return`s that I placed in there with `panic`s, but I am used to having an `errChan` or a error variable that the main thread is checking, logs and kills the service if it ever gets a value.

For this library those would have to be global variables, or passed to every "child" of `Central`, but I don't think this library has any of those. I am kinda leaning to wards just telling Copilot to just call panic... it feels the most straight forward.

permitchecklog.go
```diff
@@ -0,0 +1,204 @@
...
+			close(t.stopChan)
+			<-t.doneChan      // Wait for main flush management goroutine to finish
+			<-t.flushDoneChan // Wait for any ongoing flush operation to complete
+		}
+	})
+}
+
+// flush writes the in-memory logs to persistent storage and clears the memory
+// TODO : Implement a more generic persistence abstraction (e.g. Provider pattern)
+func (t *CheckUsageTracker) flush() {
```
FritzOnFire: Bad bot. This entire function is the result of using an array instead of a channel to manage the `*CheckLogEntry`s. Please remove all of this unnecessary complexity (delete the function).

permitchecklog.go
```diff
@@ -0,0 +1,204 @@
...
+			if err != nil {
+				return err
+			}
+		}
+	} else {
+		// this is for testing, NOT production
+		if t.config.Test_MemDump {
+			// If no database connection, just clear
+			for _, entry := range logs {
+				t.memDump = append(t.memDump, checkLogString(entry))
```
FritzOnFire: @bosvos I don't think I saw this `memDump` being mentioned in your prompt to Copilot. My gut is telling me that this should rather just be
```suggestion
				t.log.Debug(checkLogString(entry))
```
but I am not sure what the original requirement of this was, as the comments make it seem like this was supposed to be logged to the log file, and not stored in memory.

migrations.go
```diff
@@ -282,6 +282,9 @@ func createMigrations() []migration.Migrator {
 
 		// 16. Add authuserstats table that will record user_id, enabled_date, disabled_date, last_login_date
 		simple(`CREATE TABLE authuserstats (user_id BIGINT PRIMARY KEY, enabled_date TIMESTAMP, disabled_date TIMESTAMP, last_login_date TIMESTAMP);`),
+
+		// 17. Add: session_check_logs
+		simple(`CREATE TABLE session_check_logs (id bigserial PRIMARY KEY, ts TIMESTAMP WITH TIME ZONE NOT NULL, session_token VARCHAR NOT NULL, user_id BIGINT NOT NULL);`),
```
FritzOnFire: @bosvos your prompt to Copilot did not mention how these logs are going to be read back. Which make this change to this library feel a bit half baked.

From my point of view. I can imagine three functions/use cases

```go
func (t *CheckUsageTracker) fetchLogs(start time.Time, end time.Time, limit int64, offset int64) ([]*CheckLogEntry, error) { ... }

func (t *CheckUsageTracker) fetchLogsSession(start time.Time, end time.Time, limit int64, offset int64, session string) ([]*CheckLogEntry, error) { ... }

func (t *CheckUsageTracker) fetchLogsUser(start time.Time, end time.Time, limit int64, offset int64, userID int64) ([]*CheckLogEntry, error) { ... }
```

Based on those three functions we would need the following indexes added to this nr `17` db migration.
```suggestion
		simple(`CREATE TABLE session_check_logs (id bigserial PRIMARY KEY, ts TIMESTAMP WITH TIME ZONE NOT NULL, session_token VARCHAR NOT NULL, user_id BIGINT NOT NULL);
		CREATE INDEX idx_session_check_logs_ts ON session_check_logs (ts);
		CREATE INDEX idx_session_check_logs_ts_session ON session_check_logs (ts, session_token);
		CREATE INDEX idx_session_check_logs_ts_user_id ON session_check_logs (ts, user_id);
```
BUT this would make the table quite index heavy... and while we should not need to worry about that anymore (after Copilot implements my changes to refactor the code to make use of a channel instead of an array for the log entries buffer, it will be very difficult to have the code block execution of the main thread), it still feels like this is going to become slower/more expensive as time goes on without making use of table partitioning or pruning of this table.

Either way, I think all we really need is the first index I recommended on `ts`, as that should filter out enough data to make the rest of the filtering very quick, even if needs to brute force the rest.

